PR #20, opened 2022-07-07
fix: update python version to 3.7 and requirement.txt

dreamcoin1998:
Many third parties of tapshell rely on python3.7, but the python version of docker image is python3.6, which leads to the failure of image construction.


xbsura (approved):
good but not good enough

previous pr is not available because of python version, so please make ci/cd work asap, before nexr pr merge

PR #128, opened 2022-10-14
fix(Compile): Some code was not removed caused the compile error.

dobybros:
The compile errors was not found during test run, cause it was installed locally... my fault.


tjworks (approved):
LGTM

PR #232, opened 2022-11-26
fix(MongodbConnector): Resolved mongodb sample records for dataType i…

dobybros:
…ssue.

NULL is lowest priority, if any non-NULL type existing on field, NULL will never be able to replace the non-NULL type.


xbsura (approved):
LGTM

PR #237, opened 2022-11-27
fix(MongodbConnector): No more null type in TapFields; Added test cas…

dobybros:
…e to verify unknown field can be processed properly.


xbsura (approved):
LGTM

PR #244, opened 2022-11-28
add  corn feature

ningmeng777:


manager/tm/src/main/java/com/tapdata/tm/task/service/TaskService.java
```diff
@@ -2536,6 +2537,7 @@ public void start(ObjectId id, UserDetail user) {
      *                  第二位 是否开启打点任务      1 是   0 否
      */
     private void start(TaskDto taskDto, UserDetail user) {
+        addScheduleTask(taskDto);
```
jiuyetx, line 2540: can use AOP @after， no intrusion into the original method

jiuyetx (commented):
LGTM

PR #303, opened 2022-12-13
fix: custom doris line_delimiter

xbsura:
自定义行分隔符


jarad0628 (approved):
above field, line delimiter is needed

PR #669, opened 2023-03-29
fix: #138151

jarad0628:
oracle default must before null


HarsenLin (approved):
LGTM， @jarad0628 tested MySQL Oracle SQLServer PG DB2 is ok.

PR #1123, opened 2023-07-05
fix: #147173

jarad0628:
doris with and without http://


connectors/doris-connector/src/main/java/io/tapdata/connector/doris/bean/DorisConfig.java
```diff
@@ -25,6 +25,7 @@ public DorisConfig() {
         setDbType("doris");
         setEscapeChar('`');
         setJdbcDriver("com.mysql.cj.jdbc.Driver");
+        setDorisHttp(getDorisHttp().replace("http://", ""));
```
jiuyetx, line 28: Is there a possibility of an https string

jiuyetx (approved):

PR #1170, opened 2023-07-17
fix: #147356 Join任务报错Can't find a codec for class java.time.Instant.

weiliang110100:


iengine/iengine-common/src/main/java/io/tapdata/construct/constructImpl/BytesIMap.java
```diff
@@ -29,6 +39,9 @@ private T deserialized(byte[] data) {
 
 	@Override
 	public int insert(String key, T data) throws Exception {
+		if(data instanceof HashMap) {
```
ply0011, line 42: 可以不用重复判断instanceof

PR #1448, opened 2023-10-30
Feat(disabled-node): #149108 The legacy issue with node disabling is that if a JOIN node is disabled, it can only disable the entire link. Need to detect and disable a complete link.

11000100111010101100111:


manager/tm-common/src/test/java/com/tapdata/tm/dag/check/CheckDAGTest.java
```diff
@@ -0,0 +1,287 @@
+package com.tapdata.tm.dag.check;
```
ply0011, line 1: 1. 每个场景一个单测方法
2. 根据不同的场景，设置不同的断言，而不是所有都用assertTrue

ply0011 (approved):

jackin-code (approved):

PR #1638, opened 2024-01-09
fix(python): print all message info when execute ProcessBuilder shell to avoid process blocking

11000100111010101100111:


iengine/modules/observable-module/src/main/java/io/tapdata/observable/metric/py/PythonUtils.java
```diff
@@ -196,6 +200,34 @@ protected void unPackageFile(File setUpPyFile, File afterUnzipFile, final String
         }
     }
 
+    protected void printInfo(InputStream stream, Log logger) {
+        new Thread(() -> printMsg(stream, logger)).start();
```
ply0011, line 204: 线程设置名称，固定前缀

ply0011 (changes requested):

ply0011 (approved):

jackin-code (approved):

PR #1657, opened 2024-01-12
Feat develop rm project

ljvv7:


ply0011 (changes requested):
Comfirm connectors/rm-translate-tp/../rmProject.json and pom.xml is necessary

PR #1803, opened 2024-03-29
Fix develop task service

ljvv7:


jackin-code (approved):

jackin-code:
exception case: sonarq's new code scan issue.

PR #1816, opened 2024-04-07
fix: fix for create partition read data node when task type is only cdc

mnianqi:


iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/task/impl/HazelcastTaskService.java
```diff
@@ -489,7 +485,7 @@ public static HazelcastBaseNode createNode(
 								readPartitionOptions = ((DataParentNode<?>) node).getReadPartitionOptions();
 							}
 
-							if (readPartitionOptions != null && readPartitionOptions.isEnable() && readPartitionOptions.getSplitType() != ReadPartitionOptions.SPLIT_TYPE_NONE) {
+							if (readPartitionOptions != null && readPartitionOptions.isEnable() && readPartitionOptions.getSplitType() != ReadPartitionOptions.SPLIT_TYPE_NONE && !Objects.equals(taskDto.getType(),"cdc")) {
```
ply0011, line 488: 用SyncStage枚举类

ply0011 (changes requested):

PR #1821, opened 2024-04-09
fix(iengine): Fixed the problem of dynamically adding a table and rep…

ply0011:
…orting an error that the model cannot be found.

Closes TAP-2037


jackin-code (approved):
this is a exception case, we just fixed the issus first and next sprint complete the unit test coverage.

PR #1849, opened 2024-04-15
T-10.1.1: mock to kafka 20 fields

ply0011:
The task running in initial sync only may encounter errors or data loss towards the end


jackin-code:
exception case: sonarq's bug, new code report doesn't correct

jackin-code (approved):

PR #1898, opened 2024-05-06
Fix tap 2436 sqlexcecute v3.5.12

ningmeng777:


ply0011:
Exception case: sonarqube's bug, new code issues not because of this pr

ply0011 (approved):

PR #1902, opened 2024-05-06
fix(tm): Handling the problem of slow access speed to the testing env…

ply0011:
…ironment

Creating indexes during tm initialization to solve this problem

Closes TAP-2468


jackin-code (approved):

xbsura (changes requested):
db.fs.chunks.createIndex( { files_id: 1, n: 1 }, { unique: true } );

xbsura:
for mongodb, gridfs will use files_id to do query, and use n to do sort, so { files_id: 1, n: 1 } is a better choice, and this is in mongodb document: https://www.mongodb.com/zh-cn/docs/v6.0/core/gridfs/

and also, please create index in fs.files: db.fs.files.createIndex( { filename: 1, uploadDate: 1 } );

ply0011:
MongoDB Driver will create indexs on fs.chunks, but on dev environment, this index is missing. Suspect it was manual deleted. When attempting to rebuild, a duplicate index error occurs. Close this pr is the final decision.

PR #2081, opened 2024-07-09
Tap 3769 fix deduct bug release v3.9.0

weiliang110100:


ply0011:
## Exception case
Merging into the release branch does not require a Sonar scan, starting from release-v3.10.0

ply0011 (approved):

PR #2202, opened 2024-08-19
fix:TAP-4484 update mongo driver solve Invalid BSON field name oplog.rs

weiliang110100:


ply0011 (changes requested):
能否将mongodb驱动依赖的版本统一管理，比如加到manager->pom.xml的dependencyManagement

ply0011 (approved):

PR #2277, opened 2024-10-07
feat: supports a new scheduling strategy framework.

realxujiang:
In practical scenarios, we need to strictly control the number of tasks started on each Agent to avoid overloading a single Pod, which could lead to performance issues. We have abstracted a framework that allows the use of different task scheduling strategies based on configuration in various environments. The local deployment mode is suitable for community use, while the cloud scheduling mode supports elastic task scheduling and over-commitment in a Kubernetes environment. As a result, we have abstracted a new interface design where only the core worker selection algorithm needs to be implemented. This enables influencing the scheduling results according to the new algorithm.


xbsura:
thanks, please wait for a few days

realxujiang:
> thanks, please wait for a few days

ok 🍻

PR #2408, opened 2024-12-11
optimization log

openlg:


ply0011 (approved):
## Exception case

This pr no need to write unit tests

PR #2457, opened 2025-01-09
TAP-5541 :fix(iengine): Indexes cannot be synchronized when the updat…

ljvv7:
…e condition is not a physical primary key


ply0011 (changes requested):
建议在io.tapdata.flow.engine.V2.node.hazelcast.data.pdk.HazelcastTargetPdkDataNode#createTable方法改动

ply0011 (approved):

PR #2564, opened 2025-04-15
fix: String compression for stream offset

jarad0628:


xbsura:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **73.7**


See Sonar Scan Result at: http://58.251.34.123:29000/dashboard?branch=fix-develp-string-compression&id=tapdata

xbsura:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **73.7**


See Sonar Scan Result at: http://58.251.34.123:29000/dashboard?branch=fix-develp-string-compression&id=tapdata

iengine/iengine-common/src/main/java/com/tapdata/constant/StringCompression.java
```diff
@@ -52,4 +65,20 @@ public static String uncompress(String str) throws IOException {
 		}
 		return out.toString(DEFAULT_CODE);
 	}
+
+	public static String uncompressV2(String str) throws IOException {
+		if (str == null || str.length() == 0) {
+			return str;
+		}
+		ByteArrayOutputStream out = new ByteArrayOutputStream();
+		ByteArrayInputStream in = new ByteArrayInputStream(Base64.getDecoder().decode(str));
```
robinlosey, line 74: I think it would be better defensive coding to wrap the Base64 decoding in a try block and catch the exception if the encoding is not in Base64.
Either, you could then log this situation properly or make the code flexible to handle gracefully both the case with or without Base64 encoding.  Something like this:


        byte[] bytes;
        boolean base64Decoded = false;
        
        // First, try Base64 decode
        try {
            bytes = Base64.getDecoder().decode(str);
            base64Decoded = true;
        } catch (IllegalArgumentException e) {
            // Not base64 encoded, treat as plain UTF-8 bytes
            bytes = str.getBytes(StandardCharsets.UTF_8);
        }
        
        ByteArrayInputStream in = new ByteArrayInputStream(byes);

iengine/iengine-common/src/main/java/com/tapdata/constant/StringCompression.java
```diff
@@ -52,4 +65,20 @@ public static String uncompress(String str) throws IOException {
...
+		GZIPInputStream gunzip = new GZIPInputStream(in);
```
robinlosey, line 75: The same is true for the GZIP decompress - use a try block and handle the exception.   Something like this:
// Now try GZIP decompress
        try (ByteArrayInputStream byteStream = new ByteArrayInputStream(bytes);
             GZIPInputStream gzip = new GZIPInputStream(byteStream);
             InputStreamReader reader = new InputStreamReader(gzip, StandardCharsets.UTF_8);
             BufferedReader in = new BufferedReader(reader)) {

            StringBuilder outStr = new StringBuilder();
            String line;
            while ((line = in.readLine()) != null) {
                outStr.append(line);
            }
            return outStr.toString();
        } catch (IOException e) {
            // Not GZIP compressed; assume plain text
            System.out.println("[DEBUG] Input string appears to be plain text.");
            return str;
        }

xbsura:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **56.3**


See Sonar Scan Result at: http://58.251.34.123:29000/dashboard?branch=fix-develp-string-compression&id=tapdata

iengine/iengine-common/src/main/java/com/tapdata/constant/StringCompression.java
robinlosey: This is handled already in the calling method above.  So, no changes needed for this.

iengine/iengine-common/src/main/java/com/tapdata/constant/StringCompression.java
robinlosey: Alternatively, 
public static String uncompressV2(String str) throws IOException {

should become:
public static String uncompressV2(String str) throws IOException, IllegalArgumentException  {

And then IllegalArgumentException should be caught in the calling method above.

robinlosey (changes requested):
Please resolve my earlier comments

xbsura:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **57.7**


See Sonar Scan Result at: http://58.251.34.123:29000/dashboard?branch=fix-develp-string-compression&id=tapdata

xbsura:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **57.7**


See Sonar Scan Result at: http://58.251.34.123:29000/dashboard?branch=fix-develp-string-compression&id=tapdata

robinlosey (approved):

PR #2635, opened 2025-05-22
fix: TAP-6623 remove id from set document if metadata instance alread…

mnianqi:
…y exists when insert metadata dtos


manager/tm/src/main/java/com/tapdata/tm/metadatainstance/service/MetadataInstancesServiceImpl.java
```diff
@@ -1209,6 +1210,18 @@ public int bulkSave(List<MetadataInstancesDto> insertMetaDataDtos,
         }
     }
 
+    protected void removeSetIdIfMetadataExists(MetadataInstancesEntity metadataInstance, Update update) {
+        Query query = new Query(Criteria.where("qualified_name").is(metadataInstance.getQualifiedName()));
+        List<MetadataInstancesEntity> existedMetadataInstancesEntities = repository.findAll(query);
```
ply0011, line 1215: 这里用findAll判断存在太浪费资源

ply0011 (changes requested):

ply0011 (approved):

ply0011 (approved):

PR #3134, opened 2026-04-09
Feat tap 9131 develop

weiliang110100:


manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -903,6 +903,10 @@ public ResponseMessage<List<MutiResponseMessage>> batchStart(@RequestParam("task
                                                                  HttpServletResponse response) {
 			UserDetail userDetail = getLoginUser();
 			List<ObjectId> taskObjectIds = taskIds.stream().map(MongoUtils::toObjectId).collect(Collectors.toList());
+            if(StringUtils.isBlank(syncType)) {
+                TaskDto taskDto = taskService.findByTaskId(taskObjectIds.get(0),"syncType");
+                syncType = taskDto.getSyncType();
```
ply0011: **Bug: NullPointerException** — `taskService.findByTaskId()` can return `null` (via `BaseService.findOne` → `.orElse(null)`), but the result is used without a null check. Calling `taskDto.getSyncType()` on a null reference will throw an NPE.

Confirmed by existing code in [`TaskServiceImpl.java` line 553-556](https://github.com/tapdata/tapdata/blob/c10a336d6b90afa3d374ad5aa8837a3a366aa5f5/manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskServiceImpl.java#L553-L556) which does a null check after calling `findByTaskId`:
```java
TaskDto existsTask = findByTaskId(task.getId(), "oldVersionTimezone");
if (null != existsTask) { ... }
```

This same pattern is repeated in 4 places: `batchStart` (line 907), `batchStop` (line 924), `batchDelete` (line 957), `batchRenew` (line 987). All need the same fix.

Suggested fix (apply to all 4 occurrences):
```java
if (StringUtils.isBlank(syncType) && CollectionUtils.isNotEmpty(taskObjectIds)) {
    TaskDto taskDto = taskService.findByTaskId(taskObjectIds.get(0), "syncType");
    if (taskDto != null) {
        syncType = taskDto.getSyncType();
    }
}
```

manager/tm/src/main/java/com/tapdata/tm/application/controller/ApplicationController.java
```diff
@@ -239,9 +290,10 @@ public ResponseMessage<ApplicationDto> upsertByWhere(@RequestParam("where") Stri
      */
     @Operation(summary = "Get modules that the application has permission to access")
     @GetMapping("{id}/modules")
-    public ResponseMessage<List<ModulePermissionVo>> getAccessibleModules(@PathVariable("id") String id) {
-        List<ModulePermissionVo> modules = applicationService.getAccessibleModules(id, getLoginUser());
-        return success(modules);
+    public ResponseMessage<List<ModulePermissionVo>> getAccessibleModules(HttpServletRequest request, @PathVariable("id") String id) {
+        UserDetail userDetail = getLoginUser();
+        return success(dataPermissionCheckOfId(request, userDetail, id, null,
+                () -> applicationService.getAccessibleModules(id, userDetail)));
```
ply0011, line 296: **Bug: NullPointerException** — Passing `null` as the `actionEnums` parameter will cause an NPE in the enterprise `DataPermissionHelperImpl.checkOfQuery` at [line 251](https://github.com/tapdata/tapdata/blob/c10a336d6b90afa3d374ad5aa8837a3a366aa5f5/tapdata-enterprise/tm-enterprise/src/main/java/com/tapdata/tm/permissions/DataPermissionHelperImpl.java#L251):

```java
} else if (null != dto.getPermissionActions() && dto.getPermissionActions().contains(actionEnums.name())) {
```

`actionEnums.name()` throws NPE when `actionEnums` is `null`. This is reachable when a non-admin, non-owner user accesses modules of an Application entity that has `permissionActions` set.

Since `getAccessibleModules` is a read operation, `DataPermissionActionEnums.View` should be passed instead of `null`:

```suggestion
        return success(dataPermissionCheckOfId(request, userDetail, id, DataPermissionActionEnums.View,
```

ply0011 (commented):
## Code Review

Found 2 issues while reviewing for bugs and CLAUDE.md compliance.

### Issues

1. **NullPointerException in `TaskController`** — `findByTaskId()` can return `null`, but the result is dereferenced without a null check in 4 batch methods (`batchStart`, `batchStop`, `batchDelete`, `batchRenew`).

2. **NullPointerException in `ApplicationController.getAccessibleModules`** — `null` is passed as `actionEnums` to `dataPermissionCheckOfId`, which causes an NPE in the enterprise `DataPermissionHelperImpl` when `actionEnums.name()` is called for non-admin, non-owner users.

ply0011 (approved):

PR #3139, opened 2026-04-13
Feat tap 10521 Dashboard

weiliang110100:


manager/tm/src/main/java/com/tapdata/tm/task/service/chart/ChartViewService.java
```diff
@@ -46,7 +46,7 @@ protected List<TaskDto> getViewTaskDto(UserDetail user) {
                 //共享缓存的任务设计的有点问题
                 .and("shareCache").ne(true);
         Query query = Query.query(criteria);
-        query.fields().include(TaskServiceImpl.SYNC_TYPE, TaskServiceImpl.STATUS, "statuses");
+        query.fields().include(TaskServiceImpl.SYNC_TYPE, TaskServiceImpl.STATUS, "statuses", "name","dag");
```
ply0011, line 49: **Bug: Adding `"dag"` to shared query projection causes performance regression on existing `/tasks/chart` endpoint**

`getViewTaskDto()` is shared between the existing `chart()` endpoint and the new `dashboard()` endpoint. Adding `"name"` and `"dag"` to the MongoDB field projection means the existing `/tasks/chart` endpoint will now load the full DAG structure for every task — but `chart()` only needs `syncType`, `status`, and `statuses`.

The DAG field can be very large (containing all nodes, edges, schema transformations, and connection configs). This introduces unnecessary memory pressure and increased MongoDB query time on a previously lightweight endpoint.

**Suggested fix:** Instead of modifying the shared `getViewTaskDto()`, create a separate query method (e.g. `getViewTaskDtoWithDag()`) for the dashboard use case that includes `"name"` and `"dag"`, leaving the original method unchanged.

See: https://github.com/tapdata/tapdata/blob/1be61cfe8d3f8b2622667f60be7cd5ab6e6ccbcf/manager/tm/src/main/java/com/tapdata/tm/task/service/chart/ChartViewService.java#L48-L50

manager/tm/src/main/java/com/tapdata/tm/task/service/dashboard/TaskDashboardService.java
```diff
@@ -0,0 +1,392 @@
...
+            fillThroughput(result.getSummary().getTotalThroughput(), result.getTrends().getThroughput(), trendVo, dt.needThroughputSummary(), dt.needThroughputTrend());
+        }
+        if (dt.needConnectedDbs()) {
+            fillConnectedDbs(result.getSummary().getConnectedDbs(), user);
+        }
+        if (dt.needApiRequests()) {
+            fillApiRequests(result.getSummary().getApiRequests(), result.getTrends().getApiRequests(), dashboardQuery, dt.needApiSummary(), dt.needApiTrend());
+        }
+        if (dt.needTops()) {
+            fillTops(result.getTops(), runningTasks, latestSamples, dashboardQuery.top);
```
ply0011, line 158: **Bug: Inconsistent system-task filtering between `fillTops` and throughput aggregation**

The code deliberately filters out system tasks (MDM/FDM) for throughput trend aggregation using `filterTasks`/`filterTaskIds`, but `fillTops()` receives `runningTasks` which still includes system tasks. This creates two inconsistencies:

1. A system task can appear in `topThroughputTasks` with its throughput value, but that throughput is excluded from the `totalThroughput` summary — the numbers won't add up for the user.
2. A system task could take a spot in the top-N rankings, displacing a real user task.

**Suggested fix:** Pass `filterTasks` instead of `runningTasks` to `fillTops()`, or include system tasks in throughput aggregation as well — whichever matches the intended behavior.

See: https://github.com/tapdata/tapdata/blob/1be61cfe8d3f8b2622667f60be7cd5ab6e6ccbcf/manager/tm/src/main/java/com/tapdata/tm/task/service/dashboard/TaskDashboardService.java#L156-L160

ply0011 (commented):
## Code Review

Found **2 issues**. Checked for bugs and CLAUDE.md compliance.

### Issues

1. **Performance regression on existing endpoint** — Adding `"dag"` to the shared `getViewTaskDto()` projection causes the existing `/tasks/chart` endpoint to load large DAG structures unnecessarily.

2. **Inconsistent system-task filtering** — `fillTops()` uses unfiltered `runningTasks` (includes MDM/FDM system tasks) while throughput aggregation uses `filterTaskIds` (excludes them), leading to inconsistent dashboard data.

ply0011 (approved):

PR #3140, opened 2026-04-13
feat(TAP-10699): [to dev] optimize merge/join node performance

11000100111010101100111:


iengine/iengine-common/src/main/java/com/tapdata/constant/MapUtil.java
```diff
@@ -47,68 +62,50 @@ public static Object getValueByKey(Map<String, Object> dataMap, String key) thro
...
+		if (start < len) {
+			String subKey = key.substring(start);
+			subKey = replaceIfNeeded(subKey, replacement);
+			if (current instanceof Map) {
+				return ((Map<?, ?>) current).get(subKey);
 			}
 		}
-
-		return value;
+		return null;
```
ply0011: **Bug: flat-key fallback removed from `getValueByKey`**

The old implementation had a fallback: when nested path traversal returned `null`, it retried the lookup using the full dotted key as a flat key (via `dataMap.containsKey(trimKey(key))`). That fallback is gone in this rewrite.

**Concrete regression:** for `dataMap = {"a.b.c": 42}` and `key = "a.b.c"`:
- **Old:** nested lookup fails on `"a"` → falls back to `dataMap.get("a.b.c")` → returns `42`
- **New:** at `i=1` (first dot), `subKey = "a"`, `dataMap.get("a")` is `null` → `return null` immediately (line 83), never tries the flat key

The same regression applies to Oracle-style quoted keys with embedded dots (e.g. `"\"a.b.c\""`): the old `trimKey` + flat-lookup path handled stripping quotes before the flat retry; the new code only strips quotes in the single-token branch (`firstDot <= 0`), so quoted-dotted keys that have an interior dot enter the nested loop, produce a quoted prefix as `subKey`, and miss.

[Old code that was removed](https://github.com/tapdata/tapdata/blob/2a514197e4b888a7844af9ceece1bee1bb7c176b/iengine/iengine-common/src/main/java/com/tapdata/constant/MapUtil.java#L74-L95) should have preserved a flat-key fallback after the loop. Consider adding it back:

```suggestion
		if (start < len) {
			String subKey = key.substring(start);
			subKey = replaceIfNeeded(subKey, replacement);
			if (current instanceof Map) {
				return ((Map<?, ?>) current).get(subKey);
			}
		}
		// Fallback: try the full dotted key as a flat key (handles literal-dot keys)
		String flatKey = trimAndReplace(key, replacement);
		return dataMap.get(flatKey);
	}
```

iengine/iengine-common/src/main/java/com/tapdata/constant/MapUtilV2.java
```diff
@@ -35,59 +35,43 @@ private MapUtilV2() {
...
-						value = dataMap.getOrDefault(key, notExistsNode);
-					}
-				} else if (value instanceof List) {
-					value = CollectionUtil.getValueByKey((List) value, recursiveKey);
-				} else if (value instanceof TapArrayValue) {
-					value = CollectionUtil.getValueByKey(((TapArrayValue) value).getValue(), recursiveKey);
+				String subKey = key.substring(start, i);
+				current = parse(current, subKey);
+				if (null == current) {
+					return notExistsNode;
```
ply0011: **Bug: `null` value and absent key now treated identically, breaking the documented contract**

The class Javadoc at line 35 explicitly states:
> 注意：使用时,如果不存在会返回`NotExistsNode`,这个类用于区分不存在还是Null

The old implementation used `map.getOrDefault(subKey, notExistsNode)` to distinguish two cases:
- **key absent** → `notExistsNode`
- **key present, value is `null`** → `null`

The new `parse()` helper (line 63–64) uses `((Map<?, ?>) current).get(subKey)`, which returns `null` in **both** cases. The caller (line 53–54) then converts `null` → `notExistsNode`, collapsing the two into the same result.

**Concrete regression:** for `dataMap = {"a": null}` and `key = "a"`:
- **Old:** single-level path → `dataMap.getOrDefault("a", notExistsNode)` → returns `null` (key is present)
- **New:** `parse(dataMap, "a")` → `map.get("a")` = `null` → caller returns `notExistsNode` (incorrectly signals "absent")

Any downstream merge/join/field-processor logic that relies on the `null` return to detect a CDC SET-TO-NULL event will now mis-classify it as a missing field. This is particularly risky given this PR targets the merge/join node.

Note: The flat-key fallback removal (similar to `MapUtil`) appears intentional based on updated tests, but this null-vs-absent semantic change has no new test coverage and directly violates the Javadoc contract that remains unchanged.

**Fix:** use `containsKey` before `get` in `parse()` to preserve the distinction:
```java
static Object parse(Object current, String subKey) {
    if (current instanceof Map<?, ?> m) {
        return m.containsKey(subKey) ? m.get(subKey) : notExistsNode;
    } else if (current instanceof TapMapValue v) {
        Map<String, Object> inner = v.getValue();
        return inner.containsKey(subKey) ? inner.get(subKey) : notExistsNode;
    } else if (current instanceof List<?> v) {
        return CollectionUtil.getValueByKey(v, subKey);
    } else if (current instanceof TapArrayValue v) {
        return CollectionUtil.getValueByKey(v.getValue(), subKey);
    }
    return notExistsNode;
}
```
Then the caller's `if (null == current) return notExistsNode` check becomes unnecessary (since `parse` now only returns `null` when the container itself held a `null` value, which is the correct "present-null" signal).

ply0011 (commented):
## Code Review

Two high-signal bugs found in the `MapUtil`/`MapUtilV2` rewrite. See inline comments for details.

**Summary of issues:**

1. **`MapUtil.getValueByKey` — flat-key fallback removed** (`MapUtil.java` lines 74–95): The old code retried with the full dotted key as a flat key when nested traversal failed (to support maps whose keys contain literal `.` characters). This fallback is gone; callers that store literal-dot keys will silently get `null` where they previously got the stored value.

2. **`MapUtilV2.getValueByKey` — `null` value and absent key now indistinguishable** (`MapUtilV2.java` lines 51–54): The new `parse()` helper uses `map.get(subKey)` which returns `null` for both a present-but-null value and an absent key. The caller then converts `null` → `notExistsNode`. This violates the class's documented contract (*"用于区分不存在还是Null"*) and is untested. Merge/join CDC paths that distinguish SET-TO-NULL from MISSING-FIELD will be affected.

PR #3146, opened 2026-04-15
fix: TAP-10848 reslove Multi-node shared cache inconsistency

mnianqi:


HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **2**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-10848&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **5**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-10848&id=tapdata

iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java
```diff
@@ -0,0 +1,197 @@
...
+                String mapName = inv.getString("mapName");
+                String key = inv.getString("key");
+                
+                if (mapName != null && key != null) {
+                    evictCache(mapName, key);
+                }
+            }
+            
+            // 更新最后检查时间
+            lastCheckTimestamp.set(now);
```
ply0011, line 177: **Lost-event race on the watermark advance.**

`now` is sampled *before* the Mongo `find` runs, and then `lastCheckTimestamp.set(now)` unconditionally advances the watermark after processing. Any invalidation document whose `timestamp` is `≤ now` but which only becomes visible to the query after the cursor has already snapshotted (e.g. a concurrent `insertOne` during the `find`/`into` traversal) will be permanently skipped on the next poll — which uses `Filters.gt("timestamp", now)`.

Under clock skew across engine nodes this is even worse: a remote event stamped slightly behind this node's `now` can never be picked up.

Recommended fix: advance the watermark only to the **maximum `timestamp` actually observed** in the returned batch (or use `>=` with per-document dedup), and sample the watermark *after* reading — never from a pre-query wall-clock.

See https://github.com/tapdata/tapdata/blob/347eeaaff5b97cec895cef8a4a17061111e3eeb6/iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java#L149-L178.

iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java
```diff
@@ -0,0 +1,197 @@
...
+    private void checkAndEvictCache() {
+        try {
+            long lastCheck = lastCheckTimestamp.get();
+            long now = System.currentTimeMillis();
+            
+            // 查询新的失效事件
+            List<Document> invalidations = mongoClient.getDatabase(databaseName)
+                .getCollection(COLLECTION_NAME)
+                .find(Filters.gt("timestamp", lastCheck))
+                .limit(BATCH_SIZE)
```
ply0011, line 157: **Silent truncation when more than `BATCH_SIZE` events arrive in a window.**

`.find(Filters.gt("timestamp", lastCheck)).limit(BATCH_SIZE)` is issued with **no sort**. If more than 1000 invalidation events land within the 5s poll interval, Mongo returns an arbitrary 1000, and the subsequent `lastCheckTimestamp.set(now)` advances the watermark past the untouched remainder — those events are dropped forever.

Recommended fix: add `.sort(Indexes.ascending("timestamp"))`, advance the watermark only to the last processed document's `timestamp`, and loop until the page is smaller than `BATCH_SIZE`.

See https://github.com/tapdata/tapdata/blob/347eeaaff5b97cec895cef8a4a17061111e3eeb6/iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java#L153-L161.

iengine/iengine-common/src/main/java/com/tapdata/constant/HazelcastUtil.java
```diff
@@ -209,4 +217,47 @@ public static List<Stage> node2Stages(Node node) {
...
+	 * @param hazelcastInstance Hazelcast 实例
+	 * @param mongoUri MongoDB 连接 URI
+	 * @param databaseName MongoDB 数据库名
+	 */
+	public static void initCacheInvalidationService(HazelcastInstance hazelcastInstance, String mongoUri, String databaseName) {
+		if (cacheInvalidationService == null) {
+			synchronized (CACHE_INVALIDATION_LOCK) {
+				if (cacheInvalidationService == null) {
+					try {
+						MongoClient mongoClient = MongoClients.create(mongoUri);
```
ply0011, line 234: **`MongoClient` leak — the created client is never closed.**

`MongoClients.create(mongoUri)` here allocates a dedicated client (with its own connection pool) and hands it to `CacheInvalidationService`. However `CacheInvalidationService.stop()` only shuts down the scheduler — it never calls `mongoClient.close()`. `shutdownCacheInvalidationService` likewise drops the static reference without closing the client. Every engine restart / re-init leaks a full Mongo connection pool.

Recommended fix: either close the client inside `CacheInvalidationService.stop()` (document ownership there), or reuse the existing `clientMongoOperator` MongoClient instead of creating a second one.

See https://github.com/tapdata/tapdata/blob/347eeaaff5b97cec895cef8a4a17061111e3eeb6/iengine/iengine-common/src/main/java/com/tapdata/constant/HazelcastUtil.java#L232-L246.

ply0011 (commented):
## Code review

Found 3 high-signal issues in the new cross-cluster cache invalidation path. All three affect correctness of the PR's stated goal (reliable propagation of invalidations).

1. Lost-event race on the watermark advance in `checkAndEvictCache`.
2. Silent truncation when more than `BATCH_SIZE` events arrive in one poll window (no sort on the query).
3. `MongoClient` created in `HazelcastUtil.initCacheInvalidationService` is never closed — connection pool leak on every restart.

See inline comments for details.

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **5**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-10848&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **2**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-10848&id=tapdata

iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java
```diff
@@ -0,0 +1,239 @@
...
+     * @param mapName IMap 名称
+     * @param key 失效的 key
+     */
+    public void publishInvalidation(String mapName, String key) {
+        try {
+            Document invalidation = new Document()
+                .append("mapName", mapName)
+                .append("key", key)
+                .append("timestamp", System.currentTimeMillis())
+                .append("ttl", new Date(System.currentTimeMillis() + 60000)); // 1 分钟后自动删除
```
ply0011, line 147: **Self-eviction: publisher consumes its own invalidation events**

The published `Document` contains only `mapName` / `key` / `timestamp` / `ttl` — no `nodeId` / `instanceId` or any source identifier — and `checkAndEvictCache` filters only by `Filters.gte("timestamp", lastCheck)`, with no source filter. As a result, the same node's 5-second polling thread fetches its own document and calls `IMap.evict(...)` on the entry it just wrote in [`HazelcastTargetPdkCacheNode#processEvents`](https://github.com/tapdata/tapdata/blob/b56818d61ce3fe2a553dfd343e4a9a7916df103b/iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkCacheNode.java#L94-L107). The in-memory `processedEventIds` set only suppresses duplicates within a single process — it does not prevent the publisher from reacting to its own first event.

Effect: every successful write on a publishing node is evicted from its own local cache within ~5 seconds, defeating the cache for the writer.

Suggested fix: include a `nodeId` (e.g., `hazelcastInstance.getCluster().getLocalMember().getUuid()`) in the published document, and have `checkAndEvictCache` skip events whose `nodeId` matches the local node.

iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java
```diff
@@ -0,0 +1,239 @@
...
+            int totalProcessed = 0;
+            boolean hasMore = true;
+
+            while (hasMore && totalProcessed < BATCH_SIZE * 10) {
+                List<Document> invalidations = mongoClient.getDatabase(databaseName)
+                    .getCollection(COLLECTION_NAME)
+                    .find(Filters.gte("timestamp", lastCheck))
+                    .sort(Sorts.ascending("timestamp"))
+                    .limit(BATCH_SIZE)
+                    .into(new ArrayList<>());
```
ply0011: **Cursor / timestamp issues in `checkAndEvictCache`**

Three issues compound here:

1. **Inner loop makes no forward progress.** `lastCheck` is captured once before the `while` and never re-read inside, so when ≥ `BATCH_SIZE` documents share the same millisecond, every iteration of the `while` re-runs the same `find(...)` and returns the same prefix. Termination relies on the `BATCH_SIZE * 10` cap, costing up to ~10 redundant Mongo round-trips per tick (the `processedEventIds` set masks any double-eviction but the work is wasted).

2. **`Filters.gte` re-fetches the boundary timestamp every tick** because `lastCheckTimestamp.set(maxTimestamp)` advances to the max-seen timestamp. The boundary events are then re-read on the next tick and only the in-memory dedup set prevents redundant `IMap.evict`.

3. **Cross-node clock skew silently drops events (correctness bug).** `timestamp` is generated client-side via `System.currentTimeMillis()` in [`publishInvalidation`](https://github.com/tapdata/tapdata/blob/b56818d61ce3fe2a553dfd343e4a9a7916df103b/iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java#L135-L151). If node A (clock fast) publishes at `t=1000005` and the subscriber advances `lastCheckTimestamp` to `1000005`, then node B (slower clock) publishing at `t=1000002` is permanently dropped — the next `gte(1000005)` query never matches it, and the 60-second TTL purges it before any recovery is possible. This is a fragile foundation for the multi-node consistency this PR is intended to provide; it should not depend on tight wall-clock synchronization between engine hosts.

Suggested fix: switch to MongoDB Change Streams (resume tokens give monotonic, server-ordered cursoring with no client-clock dependency), or at minimum cursor by `_id` (`ObjectId` is server-side monotonic) and use `Filters.gt` on the cursor.

iengine/iengine-common/src/main/java/com/tapdata/constant/HazelcastUtil.java
```diff
@@ -209,4 +217,43 @@ public static List<Stage> node2Stages(Node node) {
...
+	public static void shutdownCacheInvalidationService() {
+		if (cacheInvalidationService != null) {
+			synchronized (CACHE_INVALIDATION_LOCK) {
+				if (cacheInvalidationService != null) {
+					cacheInvalidationService.stop();
+					cacheInvalidationService = null;
+					logger.info("Cache invalidation service stopped");
+				}
+			}
+		}
```
ply0011, line 266: **Inconsistent monitor between `init` and `shutdown` — singleton is not protected**

[`initCacheInvalidationService`](https://github.com/tapdata/tapdata/blob/b56818d61ce3fe2a553dfd343e4a9a7916df103b/iengine/iengine-common/src/main/java/com/tapdata/constant/HazelcastUtil.java#L226-L243) is `public static synchronized`, locking `HazelcastUtil.class`. `shutdownCacheInvalidationService` synchronizes on a different monitor — the private `CACHE_INVALIDATION_LOCK` object. The two methods do **not** mutually exclude.

Concurrent init + shutdown can:
- shut down a service mid-`start()`,
- leak a `MongoClient` / scheduler thread,
- leave `cacheInvalidationService` in a half-initialized state.

Additionally, the outer `if (cacheInvalidationService != null)` check on line 249 reads the static field outside any monitor, and the field is not declared `volatile`, so cross-thread visibility is not guaranteed.

Suggested fix: use the same monitor for both methods (either both `static synchronized`, or both `synchronized (CACHE_INVALIDATION_LOCK)` with the field declared `volatile`).

iengine/iengine-common/src/test/java/com/tapdata/cache/CacheInvalidationServiceTest.java
```diff
@@ -0,0 +1,162 @@
...
+            when(imap.evict(anyString())).thenThrow(new RuntimeException("Evict error"));
+
+            FindIterable<Document> findIterable = mock(FindIterable.class);
+            when(mongoCollection.find(any(Document.class))).thenReturn(findIterable);
+            when(findIterable.sort(any(Document.class))).thenReturn(findIterable);
+
+            Document invalidationDoc = new Document()
+                .append("mapName", "testMap")
+                .append("cacheKey", "testKey")
+                .append("timestamp", new java.util.Date());
```
ply0011: **`testProcessInvalidationsWithEvictError` does not exercise the eviction path it claims to test**

The stubbed `Document` uses field names and types that do not match what production reads in [`CacheInvalidationService#checkAndEvictCache`](https://github.com/tapdata/tapdata/blob/b56818d61ce3fe2a553dfd343e4a9a7916df103b/iengine/iengine-common/src/main/java/com/tapdata/cache/CacheInvalidationService.java#L179-L185):

| | Production reads | Test sets |
|---|---|---|
| Cache key field | `inv.getString("key")` | `.append("cacheKey", ...)` |
| Timestamp type | `inv.getLong("timestamp")` (`Long`) | `new java.util.Date()` |

Consequences:
- `inv.getString("key")` returns `null`, so the `mapName != null && key != null && timestamp != null` guard short-circuits and `evictCache` is never called — the "evict error" path is never exercised.
- Even with the right field name, `inv.getLong("timestamp")` against a stored `Date` would throw `ClassCastException`, also bypassing eviction.
- The Mockito stubs on `mongoCollection.find(any(Document.class))` and `findIterable.sort(any(Document.class))` cannot match production, which passes `Filters.gte(...)` / `Sorts.ascending(...)` (these are `Bson`, not `Document`). Production also uses `.into(new ArrayList<>())`, not `.iterator()`, so the `MongoCursor` stub is dead code.

The test passes vacuously and provides false confidence. Suggested fix: align the field names/types with production (`"key"` + `Long`), use `any(Bson.class)` matchers, stub `.limit(...).into(...)` to return the list, or refactor to call the eviction logic directly.

ply0011 (commented):
## Code review

Found 4 high-signal issues, plus no CLAUDE.md compliance violations. The most impactful are:

- **Self-eviction**: every node consumes its own invalidation events because the published document carries no source identifier. Each successful write on a publisher is evicted from its own local cache within ~5 s.
- **Cursor / clock-skew correctness**: the polling cursor is keyed on client-side `System.currentTimeMillis()` with `Filters.gte`, so cross-node clock skew silently and permanently drops invalidations — which directly defeats the multi-node consistency the PR aims to provide. Combined with a non-progressing inner loop and boundary re-fetch, the polling logic also wastes a noticeable amount of work per tick.
- **Singleton lifecycle**: `init` and `shutdown` use different monitors, so the pair is not mutually exclusive; the field is also non-`volatile`.
- **Test gap**: `testProcessInvalidationsWithEvictError` does not exercise the eviction error path it claims (wrong document field name and wrong timestamp type, plus `any(Document.class)` matchers that don't match `Bson` arguments).

See inline comments for details and suggested fixes. Recommend cursoring via MongoDB Change Streams (resume tokens) or `_id`/`Filters.gt` to address both the self-eviction and clock-skew issues at the source.

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **2**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-10848&id=tapdata

ply0011 (approved):

PR #3149, opened 2026-04-16
fix(api-call): 【to dev】Initialize the optimized index list and rebuild the index

11000100111010101100111:


manager/tm/src/main/java/com/tapdata/tm/apiCalls/service/ApiCallService.java
```diff
@@ -446,6 +451,9 @@ public List<ApiCallDto> save(List<ApiCallDto> saveApiCallParamList, UserDetail u
             ApiCallEntity apiCallEntity = BeanUtil.copyProperties(saveApiCallParam, ApiCallEntity.class);
             apiCallEntity.setSucceed(ApiMetricsCompressValueUtil.checkByCode(apiCallEntity.getCode(), apiCallEntity.getHttpStatus()));
             apiCallEntity.setCreateAt(new Date());
+            if (apiCallEntity.getUserInfo() != null) {
+                apiCallEntity.setCallId(String.valueOf(apiCallEntity.getUserInfo().get("clientId")));
+            }
```
ply0011, line 456: **Bug: Wrong setter called — `setCallId` should be `setClientId`**

The entire purpose of this PR is to promote `clientId` from the nested `user_info` map to a top-level field, but this code calls `setCallId(...)` (which writes to the `call_id` MongoDB field) instead of `setClientId(...)` (the newly added field).

Effects:
- All newly saved `ApiCallEntity` documents will have `clientId = null`, making the new index and the `genericFilterCriteria` query by `clientId` permanently ineffective for new records.
- The original `callId` field gets overwritten with the client ID value, corrupting it.

See the relevant fields in [`ApiCallEntity.java`](https://github.com/tapdata/tapdata/blob/c7b75202a51ae0bb521b10dadf2d64423d17ec08/manager/tm/src/main/java/com/tapdata/tm/apiCalls/entity/ApiCallEntity.java):
- `callId` → existing `call_id` field
- `clientId` → newly added top-level field (this PR)

```suggestion
            if (apiCallEntity.getUserInfo() != null) {
                apiCallEntity.setClientId(String.valueOf(apiCallEntity.getUserInfo().get("clientId")));
            }
```

manager/tm/src/main/java/com/tapdata/tm/init/patches/daas/V4_17_6_ResetApiCallIndex.java
```diff
@@ -0,0 +1,79 @@
...
+        String modulesName = MongoUtils.getCollectionName(ModulesEntity.class);
+        createTTLIndexIfNeed(mongoTemplate, modulesName, "Modules_status_1_is_deleted_1", new Document(ModulesField.STATUS.field(), 1).append(ModulesField.IS_DELETED.field(), 1), null);
+    }
+
+    protected void loadingClientId(MongoCollection<Document> collection) {
+        Document query = new Document().append("user_info.clientId", new Document().append("$exists", true));
+        Document update = new Document().append("$set", new Document().append("clientId", "$user_info.clientId"));
+        try {
+            collection.updateMany(query, update);
+        } catch (Exception e) {
```
ply0011: **Bug: `$user_info.clientId` is treated as a literal string, not a field reference — data corruption**

In the MongoDB Java driver, field reference syntax (`"$fieldName"`) inside `$set` only works when the update is passed as an **aggregation pipeline** (i.e., a `List<Document>`). When passed as a plain `Document` (as done here), the string `"$user_info.clientId"` is stored verbatim.

As a result, every migrated document will have:
```
{ "clientId": "$user_info.clientId" }   // literal string, NOT the actual value
```
instead of the actual `clientId` value from `user_info`. This silently corrupts the entire backfill migration.

The unit test does not catch this because it only verifies that `updateMany` is called with `any(Document.class)` — it does not validate the runtime MongoDB behavior.

**Fix:** Pass the update as a `List` to use the aggregation pipeline form:
```java
// Add import: java.util.List  (or java.util.Arrays)
collection.updateMany(query, List.of(update));
```

Reference: [MongoDB aggregation pipeline updates](https://www.mongodb.com/docs/manual/reference/command/update/#update-with-aggregation-pipeline)

ply0011 (changes requested):
## Code Review

Found 2 confirmed bugs — both independently flagged by two separate analysis passes.

### Issue 1 — Wrong setter in `ApiCallService.save()` (`setCallId` → `setClientId`)

The save path calls `setCallId(...)` but should call `setClientId(...)`. This means all newly saved records will have `clientId = null`, rendering the new top-level field and its index permanently useless for new data. The existing `callId` field also gets corrupted. See inline comment.

### Issue 2 — MongoDB field reference treated as literal string in `loadingClientId`

`collection.updateMany(query, update)` with a plain `Document` does not interpret `"$user_info.clientId"` as a field reference — it stores the literal string `"$user_info.clientId"`. The aggregation pipeline form `collection.updateMany(query, List.of(update))` is required. The existing unit test does not catch this. See inline comment.

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **17.4**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix-api-call-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **14.4**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix-api-call-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **14.4**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix-api-call-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **14.4**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix-api-call-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **13.9**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix-api-call-dev&id=tapdata

manager/tm/src/main/java/com/tapdata/tm/init/patches/daas/V4_17_6_ResetApiCallIndex.java
```diff
@@ -0,0 +1,91 @@
...
+        createTTLIndexIfNeed(mongoTemplate, serverUsageName, "ServerUsage_lastUpdateTime_1", new Document(ServerUsageField.LAST_UPDATE_TIME.field(), 1), null);
+        String modulesName = MongoUtils.getCollectionName(ModulesEntity.class);
+        createTTLIndexIfNeed(mongoTemplate, modulesName, "Modules_status_1_is_deleted_1", new Document(ModulesField.STATUS.field(), 1).append(ModulesField.IS_DELETED.field(), 1), null);
+    }
+
+    protected void loadingClientId(MongoCollection<Document> collection) {
+        Document query = new Document().append("user_info.clientId", new Document().append("$exists", true));
+        Document update = new Document().append("$set", List.of(new Document().append("clientId", "$user_info.clientId")));
+        try {
+            collection.updateMany(query, update);
```
ply0011, line 77: **Bug: Invalid MongoDB aggregation pipeline update syntax — migration will silently fail**

The `$set` value here is a `List`, but a standard update document requires the `$set` value to be a `Document` (object). More critically, the field path expression `"$user_info.clientId"` is only evaluated by MongoDB when the update is passed as an **aggregation pipeline** — i.e., the entire update argument must be a `List<? extends Bson>`, not a `Document` wrapping a list.

As written, the MongoDB driver uses the `updateMany(Bson filter, Bson update)` overload (single-document update), and the server rejects it because `$set`'s value is an array. The exception is swallowed by the `catch` block and only logged as a warning, so the migration **silently fails to backfill `clientId`** for all pre-existing `ApiCall` records — defeating the primary purpose of this patch.

```suggestion
        try {
            collection.updateMany(query, List.of(new Document("$set", new Document("clientId", "$user_info.clientId"))));
```

The fix passes the update as `List<Document>`, which causes the driver to call the correct `updateMany(Bson filter, List<? extends Bson> pipeline)` overload and allows MongoDB to evaluate `"$user_info.clientId"` as a field path reference.

manager/tm/src/main/java/com/tapdata/tm/apiCalls/service/ApiCallService.java
```diff
@@ -446,6 +451,9 @@ public List<ApiCallDto> save(List<ApiCallDto> saveApiCallParamList, UserDetail u
             ApiCallEntity apiCallEntity = BeanUtil.copyProperties(saveApiCallParam, ApiCallEntity.class);
             apiCallEntity.setSucceed(ApiMetricsCompressValueUtil.checkByCode(apiCallEntity.getCode(), apiCallEntity.getHttpStatus()));
             apiCallEntity.setCreateAt(new Date());
+            if (apiCallEntity.getUserInfo() != null) {
+                apiCallEntity.setClientId(String.valueOf(apiCallEntity.getUserInfo().get("clientId")));
+            }
```
ply0011, line 456: **Bug: `String.valueOf(null)` stores the literal string `"null"` in the database**

When `userInfo` is non-null but does not contain a `"clientId"` key (e.g. anonymous calls, system calls, or any auth flow that doesn't populate this field), `userInfo.get("clientId")` returns `null`. `String.valueOf(null)` is specified by the JDK to return the **4-character string `"null"`**, not a null reference.

This causes `clientId = "null"` (as a string) to be persisted in MongoDB for every such API call. This poisons the new `clientId` index introduced in `V4_17_6_ResetApiCallIndex` and corrupts downstream filtering — for example, `genericFilterCriteria` now queries `criteria.and(Tag.CLIENT_ID).is(clientId)`, which would erroneously match these `"null"`-string records.

```suggestion
            if (apiCallEntity.getUserInfo() != null) {
                Object cid = apiCallEntity.getUserInfo().get("clientId");
                if (cid != null) {
                    apiCallEntity.setClientId(cid.toString());
                }
            }
```

ply0011 (commented):
## Code Review

Found 2 bugs in the introduced code. Both are in the `clientId` backfill/save path and will cause data corruption silently.

See inline comments for details and suggested fixes.

PR #3151, opened 2026-04-18
Feat develop add localdate

jarad0628:


tapdata-test/src/main/resources/connector-configs/db2-config.json
```diff
@@ -0,0 +1,10 @@
+{
+  "host": "192.168.1.191",
+  "port": 50000,
+  "database": "SAMPLE",
+  "schema": "DB2INST1",
+  "user": "db2inst1",
+  "password": "Gotapd888",
```
ply0011, line 7: **Security: Plaintext credentials committed to public repository**

This file contains a real-looking password and internal IP addresses committed to the public `tapdata/tapdata` repo:

```
"password": "Gotapd888"
"host": "192.168.1.191"
"rawLogServerHost": "192.168.1.132"
```

Two compounding risks:

1. **Git history is permanent.** Even after a follow-up "remove credentials" commit, the password remains in git history and is already indexed by GitHub. The password should be rotated on the DB2 instance (`db2inst1@192.168.1.191`).

2. **`src/main/resources/` is packaged into the artifact.** Per the module's `pom.xml`, everything under `src/main/resources/` is bundled into `tapdata-test.jar` and uploaded to the Maven repo (`nexus.tapdata.net`). The credentials leak via two vectors: GitHub and the distributed JAR.

**Recommended fix:**
- Replace real values with placeholders (e.g. `"password": "<your_db2_password>"`) and commit a `db2-config.example.json` template instead.
- Add `tapdata-test/src/main/resources/connector-configs/*-config.json` to `.gitignore` so local test configs are never accidentally committed.
- Rotate the `Gotapd888` password on the affected instance.

iengine/iengine-common/src/main/java/com/tapdata/constant/DateUtil.java
```diff
@@ -935,4 +935,8 @@ public static MysqlYear string2MysqlYear(String yearStr) {
 		}
 		return mysqlYear;
 	}
+
+	public static Date localDate() {
+		return new Date(new Date().getTime() + TimeZone.getDefault().getRawOffset());
```
ply0011, line 940: **Bug: `localDate()` produces a semantically incorrect `Date`**

```java
return new Date(new Date().getTime() + TimeZone.getDefault().getRawOffset());
```

`java.util.Date` stores an absolute UTC epoch instant (ms since 1970-01-01T00:00:00Z). Adding the local timezone's raw offset to the epoch value does **not** "localize" the time — it shifts the underlying UTC instant forward by the offset amount. Any consumer that reads the returned `Date` as a UTC instant (JDBC `setTimestamp`, `Instant.ofEpochMilli`, Jackson serialization, etc.) will see a time that is wrong by the offset (e.g. +8 h in CST).

Additionally, `getRawOffset()` returns the timezone's standard offset and **ignores DST**, so during daylight saving time the result is off by an additional hour.

This method currently has no callers in this PR, so nothing is broken today — but the implementation will misbehave the moment it is used.

**Suggested fix:**

```suggestion
	public static Date localDate() {
		return new Date();
	}
```

`new Date()` already represents the current moment correctly. If the intent is to produce a "display-local" timestamp for a system that stores wall-clock time as if it were UTC (e.g. a database column with no timezone info), document that assumption explicitly and handle it at the persistence layer rather than in a shared utility.

ply0011 (changes requested):
## Code Review

Two issues found.

### 1. Hardcoded credentials in `db2-config.json` (security)
The new `db2-config.json` commits a real-looking password (`Gotapd888`) and internal IPs to the public repo. Because the file lives under `src/main/resources/`, it also gets bundled into the published Maven artifact. See inline comment for remediation steps.

### 2. `DateUtil.localDate()` produces a wrong `Date` (logic bug)
Adding `TimeZone.getDefault().getRawOffset()` to the UTC epoch shifts the underlying instant by the timezone offset — any downstream code reading it as UTC will be off by the timezone offset. `getRawOffset()` also ignores DST. The method has no callers yet in this PR, but it will misbehave as soon as it is used. See inline comment for the suggested fix.

ply0011 (approved):

PR #3156, opened 2026-04-22
Feat tap 11093

weiliang110100:


manager/tm/src/main/java/com/tapdata/tm/init/patches/daas/V4_18_1_CustomNode_Aggregate.java
```diff
@@ -0,0 +1,365 @@
...
+                            returnData = returnData[0]
+                            opList = opList[0]
+                          }
+                
+                          logWarn('==opList==')
+                          logWarn(opList)
+                          logWarn('==returnData==')
+                          logWarn(returnData)
+                
+                          context.opList = opList
```
ply0011, line 291: **Bug: collapsed-result branch emits the wrong DML op**

When the two fan-out entries share the same op and identical data, the collapse reassigns `opList = opList[0]` (a scalar string) and returns a single `Map`. The Java side in [`HazelcastCustomProcessor.tryProcess`](https://github.com/tapdata/tapdata/blob/aa704624d435bd333ad7e48a9cc6d78757d911a8/iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/processor/HazelcastCustomProcessor.java#L155-L170) only consults `context.op` on the single-Map branch — `context.opList` is ignored (and `getOpList` returns `null` for non-`List` anyway). Since this branch never sets `context.op`, `defaultOp` falls back to the original event's op (`"u"` for this `case 'u':` branch), so the script intends to emit a single INSERT but downstream actually receives an UPDATE.

See [`V4_18_1_CustomNode_Aggregate.java#L281-L291`](https://github.com/tapdata/tapdata/blob/aa704624d435bd333ad7e48a9cc6d78757d911a8/manager/tm/src/main/java/com/tapdata/tm/init/patches/daas/V4_18_1_CustomNode_Aggregate.java#L280-L292).

Minimal fix — set `context.op` inside the collapse block so the single-Map path honors the intended op:

```suggestion
                          if (opList[0] === opList[1] && isEqual({...returnData[0]}, {...returnData[1]})) {
                            context.op = opList[0]
                            returnData = returnData[0]
                            opList = opList[0]
                          }

                          logWarn('==opList==')
                          logWarn(opList)
                          logWarn('==returnData==')
                          logWarn(returnData)

                          context.opList = opList
```

Alternative: don't collapse — keep `returnData`/`opList` as 1-element arrays so the existing list branch on the Java side handles them correctly.

ply0011 (commented):
## Code review

Found **1 issue**. Checked for bugs and CLAUDE.md compliance (no `CLAUDE.md` files in this repo, so compliance check was a no-op).

See inline comment on `V4_18_1_CustomNode_Aggregate.java` — the new Aggregate node's JS template has a merge-collapse branch that silently emits the wrong DML op because it sets `context.opList` to a scalar string and never sets `context.op`, so the Java single-Map path falls back to the original event's op.

ply0011 (approved):

PR #3201, opened 2026-05-11
Swagger

peng-zp:
upgrade springdoc to support spring-boot 3.x so as to Swagger UI is accessable.

### DIFF
--- a/manager/tm-api/pom.xml
+++ b/manager/tm-api/pom.xml
@@ -53,6 +53,10 @@
             <artifactId>spring-webmvc</artifactId>
         </dependency>
         <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.aspectj</groupId>
             <artifactId>aspectjweaver</artifactId>
         </dependency>
@@ -99,6 +103,8 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-autoconfigure</artifactId>
         </dependency>
+        <!-- DELETE THIS DEPENDENCY - it's incompatible with Spring Boot 3.x -->
+        <!--
         <dependency>
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-common</artifactId>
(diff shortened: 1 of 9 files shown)


ply0011:
### Code review

Found 1 issue:

1. **MD5 zero-padding breaks backward compatibility with existing stored connector checksums** (bug in `PdkSourceUtils.java`)

`PdkUtil.reDownloadIfNeed` fetches the stored MD5 via `/pdk/checkMd5/v3` (which reads GridFS metadata from previous uploads) and compares it to a freshly-computed MD5 via `PdkSourceUtils.getFileMD5(theFilePath)`:

https://github.com/tapdata/tapdata/blob/348da4af8b9df47b96b83080d409025d56f26c84/iengine/iengine-common/src/main/java/io/tapdata/flow/engine/V2/util/PdkUtil.java#L123-L131

Before this PR, `calculateFileMD5` returned `bigInt.toString(16)`, which produces a 31-character string for any MD5 whose leading hex digit is zero (e.g., `98f6bcd4621d373cade4e832627b4f6`). This PR changes the return value to `String.format("%032x", bigInt)`, always producing 32 characters (e.g., `098f6bcd4621d373cade4e832627b4f6`).

https://github.com/tapdata/tapdata/blob/348da4af8b9df47b96b83080d409025d56f26c84/manager/tm-common/src/main/java/com/tapdata/tm/utils/PdkSourceUtils.java#L60-L68

Any connector JAR uploaded before this PR may have its MD5 stored in GridFS as a 31-character string. After the upgrade, the engine recalculates a 32-character MD5 for the same file, the `.equals()` comparison at line 130 of `PdkUtil.java` fails, and the engine re-downloads every such connector JAR on its first start. The issue is self-correcting (re-downloads update the stored value), but it can cause unexpected delays at startup for deployments with many connectors.

Note: commit `a9e6ea93` (TAP-10099) and `4baadd10` show this exact zero-padding issue was previously toggled — `a9e6ea93` introduced it and `4baadd10` reverted it, suggesting awareness of this interaction. The current PR re-introduces the change without updating the stored-value comparison logic.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ply0011 (approved):

PR #3231, opened 2026-05-28
feat(TAP-9242): [to dev] Support updating task blood relationships for data traceability

11000100111010101100111:


HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **1.0**
- Status: **ERROR**, MetricKey: **new_security_hotspots_reviewed**, ActualValue: **0.0**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-9242-data-trace-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_security_hotspots_reviewed**, ActualValue: **0.0**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-9242-data-trace-dev&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_security_hotspots_reviewed**, ActualValue: **0.0**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-9242-data-trace-dev&id=tapdata

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/FieldOriginalNameMapping.java
```diff
@@ -0,0 +1,485 @@
...
+            sinkTableNodes = new ArrayList<>(tableNodeById.values());
+        }
+        if (sinkTableNodes.size() == 1) {
+            return sinkTableNodes.get(0);
+        }
+        List<LineageTableNode> nonSourceType = sinkTableNodes.stream()
+                .filter(n -> null == n.getMetadata() || !"SOURCE".equalsIgnoreCase(n.getMetadata().getSourceType()))
+                .toList();
+        List<LineageTableNode> candidates = CollectionUtils.isEmpty(nonSourceType) ? sinkTableNodes : nonSourceType;
+        candidates.sort(Comparator.comparing(Node::getId, Comparator.nullsLast(String::compareTo)));
```
ply0011, line 482: 🔴 **`candidates.sort(...)` throws `UnsupportedOperationException` on an immutable list.**

`candidates` is assigned `nonSourceType` (line 481) or `sinkTableNodes` — both produced by `Stream.toList()`, which returns an **unmodifiable** list. For any lineage graph with **2+ sink (terminal) table nodes** (e.g. ALL_STREAM/DOWNSTREAM trace of a table that fans out to multiple targets), `sinkTableNodes.size() != 1`, so this line is reached and crashes. `findTaskLineageSimply` *always* calls this (via `getTargetTableUpdateFields` and `findUpdateConditionField`), so every multi-target trace request 500s.

```suggestion
        candidates = new java.util.ArrayList<>(candidates);
        candidates.sort(Comparator.comparing(Node::getId, Comparator.nullsLast(String::compareTo)));
```

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/TrackFieldFilter.java
```diff
@@ -0,0 +1,229 @@
...
+        keptNodeIds.add(targetNodeId);
+        Map<String, Map<String, String>> result = new HashMap<>();
+
+        Map<String, Map<String, String>> targetFieldToOriginNameMap = new HashMap<>();
+        Map<String, Set<String>> targetFieldsMap = new HashMap<>();
+        targetFieldsMap.put(targetNodeId, finalTargetFields);
+
+        Map<String, Node<?>> nodeMap = new HashMap<>();
+        dag.getNodes().forEach(node -> nodeMap.put(node.getId(), node));
+        while (!nodeMap.isEmpty()) {
```
ply0011: 🔴 **Infinite loop: `while(!nodeMap.isEmpty())` can never terminate.**

A node is only removed from `nodeMap` (line 105) when `targetFieldsMap.get(id) != null`. Any node that is **not an upstream ancestor of the chosen final-target** never receives a `targetFieldsMap` entry, so the `continue` at lines 92–93 skips `nodeMap.remove(id)` forever. With a multi-target / disconnected / cyclic lineage graph, the loop spins at 100% CPU indefinitely whenever `traceFilterFieldNames` is non-empty.

Fix: drive the loop over a work-queue of nodes that actually receive field requirements (BFS from the target over `upstreamNodeIds`), instead of re-scanning all `dag.getNodes()` until `nodeMap` empties.

manager/tm/src/main/java/com/tapdata/tm/trace/service/data/TraceDataService.java
```diff
@@ -0,0 +1,1109 @@
...
+        upstreamFieldMapping.forEach((fieldName, originName) -> upstreamByOrigin.putIfAbsent(originName, fieldName));
+
+        List<Map<String, Object>> filters = new ArrayList<>();
+        if(CollectionUtils.isEmpty(currentCondition.getFilters())){
+            currentCondition.getConditionKeys().forEach(key -> {
+                if (hasCurrentRecords(currentTraceValue)) {
+                    filters.addAll(rewriteFilterWithRecords(null,currentCondition.getConditionKeys(), currentTraceValue, upstreamByOrigin, currentFieldMapping, updateConditionFieldList, errors));
+                }else if(StringUtils.isNotBlank(currentCondition.getSql())){
+                    if(currentFieldMapping.containsKey(key)){
+                        buildCondition.setSql(currentCondition.getSql().replaceAll(key,currentFieldMapping.get(key)));
```
ply0011: 🟠 **`replaceAll` treats the field name as a regex and the mapping as a regex replacement — corrupts the SQL.**

`currentCondition.getSql().replaceAll(key, currentFieldMapping.get(key))`: `key` (a field name) is compiled as a **regex**, and the replacement string interprets `$`/`\` as group references.

- A field name with regex metachars (`.`, `(`, `[`, `+`, `$` — `$` is legal in Oracle identifiers) → `PatternSyntaxException`, or unintended matches.
- A mapped value containing `$`/`\` → `IndexOutOfBoundsException` / wrong substitution.
- A short field name matches *inside* longer field names elsewhere in the SQL.

Use a tokenizer/identifier-aware rewrite, or at minimum `Pattern.quote(key)` + `Matcher.quoteReplacement(value)` with word boundaries.

manager/tm/src/main/java/com/tapdata/tm/trace/service/data/TraceDataService.java
```diff
@@ -0,0 +1,1109 @@
...
+    }
+
+    private TraceQueryCondition buildTargetCondition(WideTableTraceRequest request, Node targetNode) {
+        TraceQueryCondition condition = new TraceQueryCondition();
+        condition.setConnectionId(StringUtils.defaultIfBlank(resolveConnectionId(targetNode), request.getConnectionId()));
+        condition.setTable(StringUtils.defaultIfBlank(resolveTableName(targetNode), request.getTable()));
+        if (request.getFilters() != null && StringUtils.isNotBlank(request.getFilters().getSql())) {
+            condition.setSqlMode(true);
+            condition.setSql(request.getFilters().getSql());
+            Map<String, Object> filter = JSON.parseObject(request.getFilters().getSql(), Map.class);
```
ply0011: 🟠 **The same value is treated as SQL and parsed as JSON; crashes on the intended input.**

Lines 310–313 set `request.getFilters().getSql()` as SQL (`setSqlMode(true)` / `setSql`) **and** parse it via `JSON.parseObject(..., Map.class)` then call `.keySet()`. There is no try/catch in the `traceDag → buildTargetCondition` chain.

- A real SQL filter string (the apparent point of `sqlMode`) is not valid JSON → fastjson throws `JSONException`.
- The literal `null` or a JSON array → `parseObject` returns `null`/non-Map → `filter.keySet()` throws NPE/`ClassCastException`.

Either aborts the `StreamingResponseBody`. Decide whether the field is SQL or JSON and validate/guard accordingly.

manager/tm/src/main/java/com/tapdata/tm/trace/service/data/TraceDataService.java
```diff
@@ -0,0 +1,1109 @@
...
+                String key = info.getTargetName();
+                Object value = readRecordValue(record, key);
+                if (value != null) {
+                    filter.put(originName, value);
+                }
+            });
+        }
+        if(MapUtils.isNotEmpty(filter))return filter;
+        if (MapUtils.isEmpty(sourceFilter)) {
+            sourceKeys.forEach(key -> {
```
ply0011: 🟠 **NPE: `sourceKeys.forEach(...)` when `sourceKeys` is `null`.**

The call site at line 537 passes `sourceKeys = null`. If `updateConditionFieldList` yields no filter (line 583 doesn't early-return) and `sourceFilter` is empty, this branch dereferences the null `sourceKeys`.

Trigger: `filters.conditions` containing an empty object (e.g. `[{...},{}]`) — `buildTargetCondition` adds it verbatim, so `currentCondition.getFilters()` is non-empty but contains `{}`; with no current records the empty element reaches here and NPEs, aborting the trace.

```suggestion
            if (sourceKeys != null) sourceKeys.forEach(key -> {
```

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/JoinStateSetter.java
```diff
@@ -0,0 +1,825 @@
...
+
+    protected String toTableKey(String connectionId, String tableName) {
+        if (StringUtils.isBlank(connectionId) || StringUtils.isBlank(tableName)) {
+            return null;
+        }
+        return connectionId + "_" + tableName;
+    }
+
+    void collectJoinKeys(List<MergeTableProperties> infos, Map<String, TableProperties> mergePropertiesMap, DAG taskDag, String tableType) {
+        infos.forEach(child -> {
```
ply0011, line 504: 🟠 **NPE: `collectJoinKeys` does `infos.forEach(...)` with no null guard.**

The call site (lines 96–97) passes `mergeTableNode.getMergeProperties()`, which can be `null` — proven by the explicit null/empty check at line 466 (`containsAppendMergeType`). A `MergeTableNode` configured without merge properties throws NPE here during `markJoinState`, failing the entire lineage response for that task.

```suggestion
        if (CollectionUtils.isEmpty(infos)) {
            return;
        }
        infos.forEach(child -> {
```

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/JoinStateSetter.java
```diff
@@ -0,0 +1,825 @@
...
+        if (MapUtils.isEmpty(tasks)) {
+            return;
+        }
+        for (LineageTask lineageTask : tasks.values()) {
+            if (null == lineageTask || StringUtils.isBlank(lineageTask.getId())) {
+                continue;
+            }
+            String taskId = lineageTask.getId();
+            Node<?> taskNode = lineageTask.getTaskNode();
+            String nodeId = taskNode.getId();
```
ply0011, line 293: 🟠 **NPE: `taskNode.getId()` without null-checking `getTaskNode()`.**

`mergeInfoSetter` only guards `lineageTask` and `lineageTask.getId()` (line 288), not `getTaskNode()`. Every other loop in this class guards it (e.g. lines 70–71), so `taskNode` can legitimately be `null` → `taskNode.getId()` at line 293 throws NPE, failing `markJoinState`.

```suggestion
            Node<?> taskNode = lineageTask.getTaskNode();
            if (null == taskNode || StringUtils.isBlank(taskNode.getId())) {
                continue;
            }
            String nodeId = taskNode.getId();
```

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/TableUpdateFieldGetter.java
```diff
@@ -0,0 +1,116 @@
...
+        if (CollectionUtils.isNotEmpty(targets)) {
+            for (Node<?> node : targets) {
+                if (!(node instanceof TableNode tableNode)) {
+                    continue;
+                }
+                if (connectionId.equals(tableNode.getConnectionId()) && tableName.equals(tableNode.getTableName())) {
+                    return tableNode;
+                }
+            }
+            for (Node<?> node : targets) {
```
ply0011: 🟠 **Fallback returns the FIRST target `TableNode` regardless of `connectionId`/`tableName` match → wrong table's update-condition fields.**

When the requested `(connectionId, tableName)` is not among the task's target nodes (table renamed, casing mismatch, served by a different task), the exact-match loop fails and lines 98–102 return an arbitrary other target table. `eachTaskCandidates` then reports **that** table's `updateConditionFields`/PK as the lineage target's update-condition fields — silently wrong data. Return `null` instead of an unrelated node when no exact match is found.

manager/tm/src/main/java/com/tapdata/tm/trace/service/log/ChangeLogQuery.java
```diff
@@ -0,0 +1,197 @@
...
+        for (Node<?> node : nodes) {
+            Map<String, Object> attrInfo = Optional.ofNullable(node.getAttrs()).orElse(new HashMap<>());
+            attrInfo.put("changeLogQueryEnable", false);
+            if (node instanceof LineageTableNode tableNode) {
+                String connectionId = tableNode.getConnectionId();
+                String tableName = tableNode.getTable();
+                Criteria sub = Criteria.where("connectionId").is(connectionId)
+                        .and("tableName").is(tableName);
+                criteriaList.add(sub);
+                attrInfoMap.put(String.format("%s.%s", connectionId, tableName), attrInfo);
```
ply0011, line 176: 🟠 **`attrInfoMap` keyed by `connectionId.tableName` → only the last duplicate node gets `changeLogQueryEnable=true`.**

When one physical table appears in multiple lineage nodes (same source table feeding two tasks, or appearing as both source and target — duplicates are expected; cf. `initTableNode`'s `putIfAbsent`), the `put` at line 176 keeps only the last node's `attrInfo`. When the share-CDC mapping is found (lines 190–194), only that one node flips to enabled; the other node(s) keep `changeLogQueryEnable=false` + the "shared mining not enabled" message, wrongly telling the UI change-log is unavailable for them. Key by node id, or aggregate all attrInfos sharing a table key.

manager/tm/src/main/java/com/tapdata/tm/trace/service/bloodline/JoinStateSetter.java
```diff
@@ -0,0 +1,825 @@
...
+            List<Node> successors = taskDag.getTarget(current);
+            if (CollectionUtils.isEmpty(successors)) {
+                continue;
+            }
+            for (Node<?> succ : successors) {
+                if (null == succ || StringUtils.isBlank(succ.getId()) || !visited.add(succ.getId())) {
+                    continue;
+                }
+                if (succ instanceof JoinProcessorNode joinProcessorNode) {
+                    return before ? beforeNode : joinProcessorNode;
```
ply0011, line 383: 🟡 **`findFirstDownstreamJoinNode(..., true)` returns a BFS-global `beforeNode`, not the actual predecessor of the join on the path from the source.**

`beforeNode` is overwritten each BFS iteration across *different* branches, so in a multi-branch DAG it can point to a node on an unrelated branch right before the JOIN is discovered. It is then used as `firstBeforeJoinNode.getId()` at line 155 to resolve join-key original field names, querying the wrong node's metadata and producing incorrect `joinKeys`. Track the predecessor per-path instead of as a single shared variable.

manager/tm/src/main/java/com/tapdata/tm/lineage/analyzer/impl/TableAnalyzerV1.java
```diff
@@ -330,31 +338,27 @@ private List<TaskEntity> findTasks(String connectionId, String table, Set<String
...
 	@NotNull
-	private static Criteria buildTaskCriteria(String connectionId, String table) {
+	protected static Criteria buildTaskCriteria(String connectionId, String table) {
 		Criteria syncTaskCriteria = new Criteria("dag.nodes.connectionId").is(connectionId).and("dag.nodes.tableName").is(table);
 		Criteria migrateSrcCriteria = new Criteria("dag.nodes.tableNames").is(table);
 		Criteria migrateTgtCriteria = new Criteria("dag.nodes.syncObjects.objectNames").is(table);
 		Criteria migrateCriteria = new Criteria("dag.nodes.connectionId").is(connectionId)
 				.andOperator(new Criteria().orOperator(migrateSrcCriteria, migrateTgtCriteria));
-		Criteria notDeleteCriteria = new Criteria("is_deleted").is(false);
+		Criteria notDeleteCriteria = TaskFilter.filter(new Criteria("is_deleted").is(false));
```
ply0011, line 355: 🟡 **`TaskFilter.filter(...)` narrows lineage to exclude log-collector / conn-heartbeat / share-cache / deleting tasks — and this also affects the *existing* lineage feature.**

The old criteria was just `is_deleted == false`. `TaskFilter.filter` additionally excludes `syncType ∈ {LOG_COLLECTOR, CONN_HEARTBEAT}`, `status ∈ {DELETE_FAILED, DELETING}`, and `shareCache == true`. `buildTaskCriteria` backs both the new bloodline **and** the pre-existing lineage (`LineageService` is `@Qualifier("tableAnalyzerV1")`), so a table participating only in a shared-mining/heartbeat/share-cache task is now silently dropped from lineage that previously showed it. Looks intentional (the Chinese comments say "filter out mining/share-cache tasks"), but please confirm this behavior change to the existing feature is desired.

iengine/iengine-app/src/main/java/io/tapdata/services/ShareCdcLogContentResolver.java
```diff
@@ -0,0 +1,395 @@
...
+		return Long.parseLong(text);
+	}
+
+	private ChangeLogData readFromMongo(String ringBuffer, ExternalStorageDto externalStorage, org.bson.conversions.Bson filter, int limit, TapTable tapTable){
+		ChangeLogData changeLogData = new ChangeLogData();
+		List<Map<String, Object>> logs = new ArrayList<>();
+        try (MongoDBResource mongoDBResource = new MongoDBResource()) {
+            externalStorage.setTable(ringBuffer);
+            mongoDBResource.doInit(ExternalStorageUtil.getMongoDBConfig(externalStorage, ConstructType.IMAP, ringBuffer));
+            for (Document document : mongoDBResource.getMongoCollection().find(filter).limit(limit)) {
```
ply0011: 🟠 **Paginated change-log read has no `.sort()`, but pagination relies on key order.**

`...find(filter).limit(limit)` returns documents in natural (storage) order, yet `lastKey` is taken from the **last** document (line 231) and the next page filters `gt("key", lastKey)` (line 77). If the collection does not guarantee key-ordered natural order, `lastKey` is an arbitrary document's key rather than the page maximum, so the next page silently **skips** records between and may **re-read** others.

```suggestion
            for (Document document : mongoDBResource.getMongoCollection().find(filter).sort(com.mongodb.client.model.Sorts.ascending("key")).limit(limit)) {
```

manager/tm/src/main/java/com/tapdata/tm/task/service/TaskServiceImpl.java
```diff
@@ -3102,7 +3083,7 @@ public byte[] callEngineRpc(String engineId, String cls, String method, Object[]
     public <T> T callEngineRpc(String engineId, Class<T> returnClz, String className, String method, Object... args) throws Throwable {
         String callId = UUID.randomUUID().toString().replace("-", "");
         ServiceCaller serviceCaller = ServiceCaller.create(callId).className(className).method(method);
-        Optional.ofNullable(engineId).ifPresent(id -> serviceCaller.subscribeIds("processId_" + id));
+        serviceCaller.subscribeIds("processId_" + resolveEngineRpcEngineId(engineId));
```
ply0011: 🟡 **`callEngineRpc` changed null/blank `engineId` semantics (both overloads, also line 3039).**

Old: `Optional.ofNullable(engineId).ifPresent(id -> subscribeIds(...))` — a null/blank engineId meant *no subscription*. New: `subscribeIds("processId_" + resolveEngineRpcEngineId(engineId))` picks a **random** available agent, or throws `BizException(AGENT_NOT_FOUND)` when none are registered. Existing callers that pass an unassigned task's blank `agentId` (e.g. `mergeCacheManager` line 5924, `checkTaskMemoryHeap` line 6121) now throw `AGENT_NOT_FOUND` instead of the prior behavior, or run on one arbitrary engine. Confirm no existing caller relied on the no-subscription path.

manager/tm/src/main/resources/messages.properties
```diff
@@ -452,7 +452,15 @@ api.id.empty=API path must be not empty
...
 schema.field.name.empty=Field name can not be empty
 schema.field.type.empty=Field type can not be empty
 schema.field.type.noSupport=Unsupported field type: <{0}>, only support: {1}
+data.trace.log.sTime=Start time can not be empty
+data.trace.log.eTime=End time can not be empty
+data.trace.log.time.too.large=The query time range is too large, exceeding {0} days
+data.trace.log.supported=We currently do not support using [{0}] as a shared cache external storage configuration to read change logs
+
+data.trace.findDag.error=Unable to find dag info by: {0}, error: {1}
+data.trace.log.unsupported=Table {0} of data source connection {1} has not enabled shared mining, therefore it is unable to obtain the corresponding event change log records
```
ply0011: 🟡 **The 7 `Trace.*` message keys are missing from the base/fallback `messages.properties`.**

The base bundle received only the `data.trace.*` keys; `Trace.LineageDag.NotFound`, `Trace.TargetNode.NotFound`, `Trace.Field.Mapping.NotFound`, `Trace.Value.Mismatch`, `Trace.Filter.NotBuilt`, `Trace.Merge.JoinKey.NotFound`, and `Trace.LineageDag.NodesNotFound` exist only in `messages_en_US/zh_CN/zh_TW`. `TraceDataService` throws `BizException` with these codes — for any locale not covered by the three variant bundles (the base bundle is the ultimate fallback), the user sees the raw code string instead of a translated message. Add the keys to the base `messages.properties`.

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **3**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-9242-data-trace-dev&id=tapdata

ply0011 (approved):

PR #3236, opened 2026-06-03
fix(TAP-11874): [to dev] query settings by filter shoule use http param

11000100111010101100111:


manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        if (orderJsonObject == null) {
+            return sort;
+        }
+        List<Field> fieldLis = Arrays.asList(clazz.getDeclaredFields());
+        Sort sorted = null;
+        for (Object sortItem : orderJsonObject) {
+            for (java.lang.reflect.Field field : fieldLis) {
+                String orderByString = sortItem.toString();
+                String filedName = field.getName();
+                if (!orderByString.contains(filedName)) {
```
ply0011, line 49: **Substring field matching produces wrong/extra sort keys.** `orderByString.contains(filedName)` matches when a declared field name is a *substring* of the order string, not an exact match. With `order=["category_sort DESC"]`, the `Settings` fields `category`, `sort`, **and** `category_sort` all match, producing sort doc `{category:-1, category_sort:-1, sort:-1}` — so the primary sort key becomes `category`, not the requested `category_sort`, and results come back in the wrong order.

Use exact-name parsing. `MongoUtils.applySort(query, filter.getSort())` (same package) already splits each entry on whitespace and uses `str[0]` as the field name verbatim.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+                    continue;
+                }
+                Sort item = null;
+                if (orderByString.contains("ASC")) {
```
ply0011, line 53: **Direction detection is case-sensitive.** `contains("ASC")` / `contains("DESC")` only matches uppercase, but the documented `Filter` sort format is lowercase (`@Parameter`: "Sorting criteria in the format: property( asc|desc)"). An order like `["key asc"]` matches neither branch, `item` stays `null`, no criteria is added, and it silently falls back to the default sort — the user's requested order is dropped.

`MongoUtils.applySort` uses `Sort.Direction.fromOptionalString(...)`, which is case-insensitive; reusing it avoids this.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        JSONObject jsonObject = JSONUtil.parseObj(filter);
+        JSONObject whereJsonObject = jsonObject.getJSONObject("where");
+        Sort sort = buildSort(jsonObject, clazz);
+        Query query =  buildQuery(whereJsonObject);
+        query.with(sort);
+        return query;
+    }
+
+    public static Sort buildSort(JSONObject jsonObject, Class<?> clazz) {
+        Sort sort = Sort.by("createAt").descending();
```
ply0011, line 35: **Default sort targets a field that doesn't exist on the only caller.** `Sort.by("createAt")` assumes a `createAt` field, but `Settings` (the sole caller via `SettingsServiceImpl`) is a plain `@Data` class — it does **not** extend `BaseEntity` and has no `createAt` (it has `last_update`). MongoDB treats the missing field as null for every document, so the "default ordering" is effectively no ordering.

Prefer `MongoUtils.applySort(query, filter.getSort())`, which orders by the caller-supplied sort rather than a hard-coded field name.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        if (jsonObject == null) {
+            return sort;
+        }
+        JSONArray orderJsonObject = jsonObject.getJSONArray("order");
+        if (orderJsonObject == null) {
+            return sort;
+        }
+        List<Field> fieldLis = Arrays.asList(clazz.getDeclaredFields());
```
ply0011, line 43: **`getDeclaredFields()` excludes inherited fields.** It returns only fields declared on `clazz` itself, so for any `BaseEntity`-derived entity, sorting by inherited fields (`id`, `createAt`, `lastUpdAt`) is silently dropped — and it's internally inconsistent with the hard-coded `createAt` default on line 35 (which `getDeclaredFields()` would itself reject as a sort target). `MongoUtils.applySort` doesn't reflect over fields at all; it applies the requested field name directly.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+            for (Map.Entry<String, Object> entry : whereJsonObject.entrySet()) {
+                String proName = entry.getKey();
+                String proValue = entry.getValue().toString();
+                if (proValue.contains("$in")) {
+                    JSONObject jsonObjectValue = JSONUtil.parseObj(proValue);
+                    JSONArray jsonArray = jsonObjectValue.getJSONArray("$in");
+                    List<?> list = jsonArray.toList(Object.class);
+                    query.addCriteria(Criteria.where(proName).in(list));
+                } else {
+                    query.addCriteria(Criteria.where(proName).is(proValue));
```
ply0011, line 82: **All `where` values are coerced to String.** `entry.getValue().toString()` + `Criteria.where(proName).is(proValue)` means non-string equality and every operator except `$in` won't match. `where={"user_visible": true}` builds `is("true")` which never matches the boolean field; `where={"sort": {"$gt": 5}}` becomes a literal-string match. This branch now feeds the Settings query (previously it returned all settings unfiltered), so such filters silently return empty/wrong results.

`MongoUtils.buildCriteria(where, entityInformation)` already handles value types and Mongo operators.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+        query.with(sort);
+        return query;
+    }
+
+    public static Sort buildSort(JSONObject jsonObject, Class<?> clazz) {
+        Sort sort = Sort.by("createAt").descending();
+        if (jsonObject == null) {
+            return sort;
+        }
+        JSONArray orderJsonObject = jsonObject.getJSONArray("order");
```
ply0011, line 39: **`filter.sort` and single-String `order` are ignored.** `buildSort` reads only the JSON `order` array. `Filter.getSort()` is the canonical accessor that merges the standard `sort` field **and** a single-String `order`. So a client using `?sort=key DESC` (the documented pagination param) or `order="key DESC"` as a single string (not an array) gets `getJSONArray("order") == null` → the default sort, ignoring the request. Pass `filter.getSort()` into `MongoUtils.applySort` instead of re-reading raw JSON.

manager/tm/src/main/java/com/tapdata/tm/Settings/service/SettingsServiceImpl.java
```diff
@@ -207,7 +207,8 @@ public List<SettingsDto> findALl(String decode, Filter filter) {
                         .orElseGet(() -> findAll(query));
 
             }else {
-                settingsList = settingsRepository.findAll();
+                Query query = TapQueryUtil.buildQuery(filter, Settings.class);
```
ply0011, line 210: **Altitude / reuse — this helper duplicates existing query-building infrastructure.** `TapQueryUtil` re-implements `MongoUtils.buildCriteria` + `MongoUtils.applySort` (same `com.tapdata.tm.utils` package) and `BaseRepository.filterToQuery(filter)`, but with weaker, buggier logic — it is the root cause of the sort/where defects flagged in `TapQueryUtil.java` (substring sort matching, case-sensitive direction, missing `createAt`, type coercion, ignored `filter.getSort()`).

Consider building the query from the existing path here, e.g. `settingsRepository.filterToQuery(filter)` (or `findAll(filter)`), then `mongoTemplate.find(...)`. That path already handles operators, value types, field projection, soft-delete, and case-insensitive sort — and keeps the two settings query paths from drifting.

manager/tm-api/src/main/java/com/tapdata/tm/utils/TapQueryUtil.java
```diff
@@ -0,0 +1,89 @@
...
+ * @author <a href="2749984520@qq.com">Gavin'Xiao</a>
+ * @author <a href="https://github.com/11000100111010101100111">Gavin'Xiao</a>
+ * @version v1.0 2026/6/3 10:54 Create
+ * @description
+ */
+public final class TapQueryUtil {
+    private TapQueryUtil() {}
+
+    public static Query buildQuery(Filter filter, Class<?> clazz) {
+        JSONObject jsonObject = JSONUtil.parseObj(filter);
```
ply0011, line 26: Two issues anchored here:

1. **No null-guard on `filter`.** `BaseRepository.filterToQuery` does `if (filter == null) filter = new Filter()`; this `JSONUtil.parseObj(filter)` throws NPE on a null `Filter` instead of producing an empty (`is_deleted ne true`) query.
2. **Wasteful + lossy round-trip.** Serializing the whole `Filter` bean to JSON and re-parsing it just to read `where`/`order` runs on the hot settings-list read path, and it's what forces the lossy `toString()` String-coercion of `where` values. The typed getters `filter.getWhere()` / `filter.getSort()` are already available.

manager/tm/src/main/java/com/tapdata/tm/Settings/service/SettingsServiceImpl.java
ply0011: ⚠️ 更正:上面建议的 `settingsRepository.filterToQuery(filter)` / `findAll(filter)` **不可行**,请忽略那部分。

原因:`BaseRepository<Entity extends BaseEntity, ID>` 要求实体继承 `BaseEntity` 且面向 `ObjectId` 主键,而 `Settings` 的 `_id` 是 `String`、未继承 `BaseEntity` —— `SettingsRepository` 的注释也写明「因为 Settings 的表 _id 类型为字符串,导致这个类不能继承 BaseRepository,所以直接继承 MongoRepository」。所以这两个方法在 `SettingsRepository` 上根本不存在。

不过「复用 `MongoUtils`」这条路只有 `buildCriteria` 失效(它需要 `MappingMongoEntityInformation`,Settings 走 `MongoRepository` 拿不到现成的),而 **`MongoUtils.applySort(Query, List<String>)` 只依赖 `List<String>`,对 Settings 完全可用**,可以一次性修掉本类里所有排序相关问题(子串匹配 / 大小写 / `createAt` 默认 / 继承字段 / 忽略 `filter.getSort()`)。

可行的写法(不再需要 `TapQueryUtil`):

```java
} else {
    Query query = new Query();
    Where w = filter.getWhere();
    if (w != null) {
        w.forEach((k, v) -> {
            if (v instanceof Map && ((Map<?, ?>) v).containsKey("$in")) {
                query.addCriteria(Criteria.where(k).in((Collection<?>) ((Map<?, ?>) v).get("$in")));
            } else if (v != null) {
                query.addCriteria(Criteria.where(k).is(v)); // 保留原始类型,不再 toString()
            }
        });
    }
    query.addCriteria(Criteria.where("is_deleted").ne(true));
    MongoUtils.applySort(query, filter.getSort());          // 复用正确的排序解析
    settingsList = mongoTemplate.find(query, Settings.class);
}
```

- 排序交给 `MongoUtils.applySort` → 精确字段名匹配、大小写无关、合并 `sort`+`order`;空排序时不排序,与旧的 `findAll()` 自然序一致。
- 遍历类型化的 `Where`(本身就是 `Map<String,Object>`)→ `.is(v)` 保留布尔/数字类型,`$in` 仍支持,且无需 `JSONUtil.parseObj(filter)` 的序列化往返,也不会因 null filter 而 NPE。

若后续确实要支持 `like`/日期范围/`$oid` 等高级操作符,再考虑用 `mongoTemplate` 的 `MappingContext` 构造 `MappingMongoEntityInformation<Settings,String>` 接入 `MongoUtils.buildCriteria`;对当前仅按 category/key 过滤的场景属于过度设计。

ply0011 (approved):

PR #3266, opened 2026-06-12
Harsen/tap 11621 develop

HarsenLin:


manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
```diff
@@ -232,7 +233,52 @@ public Page<UserLogDto> find(Filter filter, UserDetail userDetail) {
...
+                    modular, operation, specificKey, defaultKey);
+            return fallbackI18nMessage(dto, modular, operation);
+        }
+
+        return MessageUtil.formatString(template, buildTemplateParams(dto, locale, modular));
+    }
+
+    private Map<String, Object> buildTemplateParams(UserLogDto dto, Locale locale, String modular) {
+        Map<String, Object> params = new HashMap<>();
+        params.put("moduleName", getModuleName(locale, modular));
```
ply0011, line 263: **[correctness] `i18nMessage` is returned only half-rendered.**

`buildTemplateParams` supplies only `moduleName`, and `MessageUtil.formatString` (Apache `StringSubstitutor`) leaves unresolved `{...}` tokens intact. So for modular=`connection`, operation=`create`, username=`alice`, parameter1=`mysql_conn` (zh_CN), `find()` returns `用户 {user} 创建了 连接 {parameter1}` instead of `用户 alice 创建了 连接 mysql_conn`.

The DTO already carries `username`/`parameter1`/`parameter2`, so if any consumer renders `i18nMessage` directly the user sees literal `{user}` tokens. The new unit tests assert this half-rendered output, locking it in. If the frontend is meant to substitute these, disregard; otherwise add them to the params map here.

manager/tm-api/src/main/java/com/tapdata/tm/config/JacksonConfig.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JacksonConfig {
+
+	@Bean
+	public SimpleModule objectIdJacksonModule() {
+		SimpleModule module = new SimpleModule();
+		module.addSerializer(ObjectId.class, new ObjectIdSerialize());
```
ply0011, line 15: **[correctness / altitude] This globally changes `ObjectId` serialization for the whole tm app, not just userLog.**

There is no `@EnableWebMvc` and no `spring.jackson` exclusion, so Spring Boot auto-registers this `SimpleModule` into the primary `ObjectMapper`. Every *unannotated* `ObjectId` field now serializes as a hex string instead of the previous default object form — e.g. `ModulesDto.connection` (returned by `ModulesController` save/update/generate/findOne/batchUpdate/getByCollectionName) and `UserLogDto.sourceId`. Any client parsing the old shape breaks.

The established convention is a one-line `@JsonSerialize(using = ObjectIdSerialize.class)` on the field (as `BaseDto.id` and ~20 other DTOs already do) — that scopes the change to what this PR actually needs. The module also registers no deserializer / key-serializer, so the round-trip is asymmetric.

manager/tm/src/main/java/com/tapdata/tm/task/service/TaskServiceImpl.java
```diff
@@ -1136,13 +1136,13 @@ public void afterRemove(TaskDto taskDto, UserDetail user) {
...
         //add message
+        Modular modular = Modular.ofTaskSyncType(taskDto.getSyncType());
         if (SyncType.MIGRATE.getValue().equals(taskDto.getSyncType())) {
             messageService.addMigration(taskDto.getDeleteName(), taskDto.getId().toString(), MsgTypeEnum.DELETED, Level.WARN, user);
-            userLogService.addUserLog(Modular.MIGRATION, Operation.DELETE, user, id.toString(), taskDto.getName());
         } else if (SyncType.SYNC.getValue().equals(taskDto.getSyncType())) {
             messageService.addSync(taskDto.getDeleteName(), taskDto.getId().toString(), MsgTypeEnum.DELETED, "", Level.WARN, user);
-            userLogService.addUserLog(Modular.SYNC, Operation.DELETE, user, id.toString(), taskDto.getName());
         }
+        userLogService.addUserLog(modular, Operation.DELETE, user, id.toString(), taskDto.getName());
```
ply0011, line 1168: **[correctness / behavior] `afterRemove` now writes a DELETE user-log for *every* sync type.**

The two per-branch `addUserLog` calls (migrate / sync) were collapsed into this single unconditional call. Deleting a `logCollector` / `connHeartbeat` / `mem_cache` task now creates an audit entry that previously did not exist, and `testRun`/`preview`/`deduceSchema` fall through `Modular.ofTaskSyncType`'s `default → MIGRATION` so they are logged as *Migration Task*.

Note the `messageService.addMigration/addSync` notification is still gated to migrate/sync only, so for the newly-logged types the audit log and the in-app message stream now diverge. Please confirm this expanded audit scope is intended.

manager/tm-api/src/main/java/com/tapdata/tm/userLog/constant/Modular.java
```diff
@@ -46,4 +51,18 @@ public static Modular of(String value) {
...
+        if (null == taskSyncType) {
+            return MIGRATION;
+        }
+        return switch (taskSyncType) {
+            case TaskDto.SYNC_TYPE_SYNC -> SYNC;
+            case TaskDto.SYNC_TYPE_MIGRATE -> MIGRATION;
+            case TaskDto.SYNC_TYPE_LOG_COLLECTOR -> LOG_COLLECTOR;
+            case TaskDto.SYNC_TYPE_CONN_HEARTBEAT -> CONN_HEARTBEAT;
+            case TaskDto.SYNC_TYPE_MEM_CACHE -> MEM_CACHE;
+            default -> MIGRATION;
```
ply0011, line 65: **[correctness] `default -> MIGRATION` mislabels every task type that isn't sync/migrate/logCollector/connHeartbeat/mem_cache.**

`testRun`, `preview`, `deduceSchema` (and `null`) all map to `MIGRATION`, so their audit messages render as *Migration Task* / *迁移任务*. Combined with the now-unconditional `addUserLog` in `afterRemove`, deleting a preview/test-run task is recorded as deleting a migration task. Consider a neutral default (e.g. `SYSTEM`) or explicit handling. The mapping test only iterates the 4 `SyncType` enum values, so this default branch is never exercised.

manager/tm-api/src/main/java/com/tapdata/tm/utils/MessageUtil.java
```diff
@@ -116,6 +116,14 @@ public static String getPdkTestItemMsg(Locale locale, String resourceId, Object.
 		return getString(locale, resourceId, msg, params);
 	}
 
+	public static String getBundleMessageOrNull(Locale locale, String bundleName, String resourceId) {
+		try {
+			return getStringOrNull(getResourceBundle(locale, bundleName), resourceId);
```
ply0011, line 123: **[correctness / i18n] `ResourceBundle.getBundle(name, locale)` falls back through the JVM default locale before the base bundle.**

Only `userLogsTemplate{,_en_US,_zh_CN,_zh_TW}.properties` ship. A request locale lacking an exact file — e.g. `Accept-Language: en` → `Locale("en")` (no `_en` file) — resolves to the JVM-default-locale candidate before the base bundle. On a server whose default locale is `zh_CN`, an English user's audit messages come back in Chinese (same for `en_GB`, `fr_FR`, `ja_JP`, …).

This is a property of the shared `getResourceBundle` helper that this PR newly routes user-facing messages through. A `ResourceBundle.Control` that skips the default-locale step (or normalizing to a supported locale) fixes it.

manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
```diff
@@ -232,7 +233,52 @@ public Page<UserLogDto> find(Filter filter, UserDetail userDetail) {
...
+        String moduleName = MessageUtil.getBundleMessageOrNull(locale, UserLogTemplateKey.BUNDLE_NAME, moduleKey);
+        if (moduleName == null) {
+            log.warn("User log i18n module name not found, modular: {}, moduleKey: {}", modular, moduleKey);
+            return modular;
+        }
+        return moduleName;
+    }
+
+    private String fallbackI18nMessage(UserLogDto dto, String modular, String operation) {
+        return valueOrEmpty(dto.getUsername()) + " " + modular + "." + operation + " " + valueOrEmpty(dto.getParameter1());
```
ply0011, line 278: **[correctness] Missing templates leak raw internal codes to the UI.**

`fallbackI18nMessage` emits `username + " " + modular + "." + operation + " " + parameter1` — e.g. `alice sync.not_exists order_sync`. It fires whenever neither a specific nor a `_default.<op>` template exists, i.e. for any stored `operation` value that isn't an `Operation` enum constant (legacy rows, future ops). Internal enum codes are shown verbatim and untranslated in zh_CN/zh_TW.

`getModuleName` (line 272) has the same failure mode, returning the raw `modular` code on a missing key. Prefer a localized generic fallback over exposing internal identifiers. The new test pins the raw output as expected.

manager/tm-api/src/test/java/com/tapdata/tm/config/JacksonConfigTest.java
```diff
@@ -0,0 +1,30 @@
...
+	void testObjectIdSerializeToHexStringInResponseMessageData() throws Exception {
+		JacksonConfig config = new JacksonConfig();
+		ObjectMapper objectMapper = new ObjectMapper();
+		ObjectId objectId = new ObjectId("66616f30c7a14d46120ef0b1");
+		Map<String, Object> data = new HashMap<>();
+		data.put("id", objectId);
+		ResponseMessage<Map<String, Object>> responseMessage = new ResponseMessage<>();
+		responseMessage.setData(data);
+
+		objectMapper.registerModule(config.objectIdJacksonModule());
```
ply0011, line 25: **[test quality] This test doesn't cover the behavior it exists to protect.**

It builds a fresh `new ObjectMapper()` and manually `registerModule(config.objectIdJacksonModule())`, so it only proves `ObjectIdSerialize` works when explicitly registered — already true before this PR. It never exercises the Spring wiring that is the whole point of `JacksonConfig`: remove `@Configuration` or `@Bean` (or add `@EnableWebMvc` / a `spring.jackson` exclusion) and global serialization silently reverts while this test stays green. It also never covers the null-`ObjectId` branch (the serializer writes nothing on null). A `@SpringBootTest`/`@WebMvcTest` asserting the auto-configured `ObjectMapper` (or an actual MVC response) would protect the real contract.

manager/tm/src/test/java/com/tapdata/tm/userLog/service/UserLogServiceImplTest.java
```diff
@@ -41,4 +56,194 @@ void testIsNull() {
...
+                        missingKeys.add(moduleKey);
+                    }
+                }
+
+                assertTrue(missingKeys.isEmpty(), templateFile + " missing module name templates: " + missingKeys);
+            }
+        }
+
+        @Test
+        void testEverySyncTypeCanMapToModular() {
```
ply0011, line 212: **[test quality] Mapping test skips the risky branch.**

`testEverySyncTypeCanMapToModular` only iterates the 4 `SyncType` enum constants, so it never covers `Modular.ofTaskSyncType`'s `default → MIGRATION` path (testRun/preview/deduceSchema/null). It gives green-test confidence that the mapping is fully correct while the mislabel of those task types (Modular.java:65) ships unguarded. Add cases for the non-enum sync strings to pin the intended behavior.

manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
```diff
@@ -232,7 +233,52 @@ public Page<UserLogDto> find(Filter filter, UserDetail userDetail) {
         long total = userLogRepository.count(filter.getWhere(), userDetail);
 
         List<UserLogDto> items = convertToDto(entityList, UserLogDto.class, "password");
+        Locale locale = MessageUtil.getLocale();
+        items.forEach(item -> item.setI18nMessage(renderI18nMessage(item, locale)));
```
ply0011, line 237: **[efficiency] Per-row exception-driven rendering on the list endpoint.**

`find()` renders `i18nMessage` for every row, and for almost every row the specific-key lookup misses (specific templates exist for only ~5 modular.operation combos), so `bundle.getString(specificKey)` throws a `MissingResourceException` that `getStringOrNull` catches before falling back to the default key. On large pages/exports that is one thrown+caught exception (stack-trace fill) per row on a hot path. Pre-checking `bundle.containsKey(...)` avoids the churn.

manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
```diff
@@ -232,7 +233,52 @@ public Page<UserLogDto> find(Filter filter, UserDetail userDetail) {
...
+        if (template == null) {
+            log.warn("User log i18n template not found, modular: {}, operation: {}, specificKey: {}, defaultKey: {}",
+                    modular, operation, specificKey, defaultKey);
+            return fallbackI18nMessage(dto, modular, operation);
+        }
+
+        return MessageUtil.formatString(template, buildTemplateParams(dto, locale, modular));
+    }
+
+    private Map<String, Object> buildTemplateParams(UserLogDto dto, Locale locale, String modular) {
```
ply0011: **[cleanup] Unused parameter.**

`buildTemplateParams(UserLogDto dto, Locale locale, String modular)` never reads `dto`; it only puts `moduleName` derived from `(locale, modular)`. Drop the param or inline the method. (If you instead address the half-rendered-message issue above by pulling `user`/`parameter1`/`parameter2` from `dto`, then `dto` becomes used — resolve the two together.)

manager/tm-api/src/main/java/com/tapdata/tm/userLog/constant/UserLogTemplateKey.java
```diff
@@ -0,0 +1,25 @@
+package com.tapdata.tm.userLog.constant;
+
+public final class UserLogTemplateKey {
+
+    public static final String BUNDLE_NAME = "userLogsTemplate";
+    public static final String PREFIX = "userLogs.";
+    public static final String INTERNAL_PREFIX = PREFIX + "_";
```
ply0011: **[cleanup] Production-dead constant.**

`INTERNAL_PREFIX` is referenced only by the test (`hasSpecificOperationTemplate`); no production code uses it. Either derive that filter locally in the test or drop the constant, so test-only concerns don't live in a production constants class.

manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
HarsenLin: 多级国际化设计：1 减少国际化的配置项，2 前端格式化处理

manager/tm-api/src/main/java/com/tapdata/tm/config/JacksonConfig.java
HarsenLin: 需要全局转换

manager/tm/src/main/java/com/tapdata/tm/task/service/TaskServiceImpl.java
HarsenLin: 预期是所有任务的操作日志都记录

manager/tm-api/src/main/java/com/tapdata/tm/userLog/constant/Modular.java
HarsenLin: 有单测验证未覆盖的任务类型

manager/tm/src/main/java/com/tapdata/tm/userLog/service/UserLogServiceImpl.java
HarsenLin: 预期设计效果

manager/tm-api/src/test/java/com/tapdata/tm/config/JacksonConfigTest.java
HarsenLin: 只需验证转换正确即可，全局转换

ply0011 (approved):

PR #3363, opened 2026-07-17
feat(TAP-12263): 【to 4.21】 Identify field mapping relationships in JS nodes

11000100111010101100111:


ply0011 (changes requested):
关闭这个pr，转为需求

PR #3370, opened 2026-07-20
feat(TAP-12097): (new feature for 4.23) support batch edit task email receivers

11000100111010101100111:


manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
         return success(taskService.checkTaskMemoryHeap(taskDto,true,getLoginUser()));
     }
 
+    @Operation(summary = "Batch modify task alarm information")
+    @PostMapping("/alarm/batch-update")
+    public ResponseMessage<List<MutiResponseMessage>> batchUpdateTaskAlarm(@RequestBody BatchUpdateAlarmParam alarm){
+        UserDetail userDetail = getLoginUser();
+        List<String> taskIds = alarm.getTaskIds();
+        List<ObjectId> taskObjectIds = taskIds.stream().map(MongoUtils::toObjectId).collect(Collectors.toList());
```
ply0011: **🔴 [correctness] NPE when `taskIds` is null (HTTP 500)**

`batchUpdateTaskAlarm` dereferences `alarm.getTaskIds()` with no null/empty guard. A request body missing `taskIds` (e.g. `{}` or `{"emailReceivers":["a@b.com"]}`) makes `getTaskIds()` return `null`, so `taskIds.stream()` throws `NullPointerException` → 500.

The sibling `batchStart`/`batchStop`/`batchDelete` endpoints avoid this by using a required `@RequestParam("taskIds")` (400 on a missing list); this `@RequestBody` field is unvalidated.

```suggestion
        if (CollectionUtils.isEmpty(taskIds)) {
            return success(new ArrayList<>());
        }
        List<ObjectId> taskObjectIds = taskIds.stream().map(MongoUtils::toObjectId).collect(Collectors.toList());
```

manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
...
+        List<String> authorizedTaskIds = new ArrayList<>();
+        List<MutiResponseMessage> noAuthMessages = new ArrayList<>();
+        for (int i = 0; i < taskIds.size(); i++) {
+            String taskId = taskIds.get(i);
+            ObjectId taskObjectId = taskObjectIds.get(i);
+            Boolean hasEditPermission = DataPermissionHelper.checkOfQuery(
+                    userDetail,
+                    DataPermissionDataTypeEnums.Task,
+                    DataPermissionActionEnums.Edit,
+                    taskService.dataPermissionFindById(taskObjectId, new Field()),
```
ply0011, line 1611: **🟠 [correctness] Malformed / unknown `taskId`s silently become `null` and bypass the error channel**

`MongoUtils.toObjectId` returns `null` (it does **not** throw) for any string that isn't 24-hex, so `taskObjectIds` can hold `null` entries that flow into `dataPermissionFindById(null, ...)`.

In the enterprise `DataPermissionHelperImpl`, `checkOfQuery` invokes the query supplier → `findById(null)` → `null` task, then `menuEnumsFun.apply(null)` calls `null.getSyncType()` → NPE, aborting the whole batch. Meanwhile the per-task `MutiResponseMessage` error channel — built precisely to report bad tasks — is never used for invalid/unknown ids. Consider validating each id up front and emitting a `MutiResponseMessage` for the invalid ones instead of forwarding `null`.

ply0011, line 1611: **🟡 [efficiency] N+1 point queries in the permission loop**

`taskService.dataPermissionFindById(...).get()` runs a `findById` once per task (plus a `new Field()` allocation each iteration). A batch of N task ids fires N sequential Mongo round-trips.

A single `find({_id: {$in: ids}})` projecting `userId`/`syncType` would replace the N queries; the existing batch endpoints already push per-task work into one service call.

manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
...
+
+        if (CollectionUtils.isNotEmpty(authorizedTaskIds)) {
+            BatchUpdateAlarmParam target = new BatchUpdateAlarmParam();
+            target.setTaskIds(authorizedTaskIds);
+            target.setAlarmSettings(alarm.getAlarmSettings());
+            target.setAlarmRules(alarm.getAlarmRules());
+            target.setEmailReceivers(alarm.getEmailReceivers());
+            alarmService.batchUpdate(target);
+        }
+        return noAuthMessages;
```
ply0011, line 1636: **🟡 [correctness] Only permission failures are returned; successes are invisible**

The endpoint returns only `noAuthMessages`. Authorized/updated tasks produce no `MutiResponseMessage`, so a caller receiving an empty list cannot distinguish "all updated" from "all silently dropped". This diverges from the `MutiResponseMessage` convention where success entries carry `code = OK` — e.g. `batchDelete` filters `responseMessages` by `ResponseMessage.OK`. There is also no per-task feedback if `alarmService.batchUpdate` partially fails.

manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
...
+            }
+        }
+
+        if (CollectionUtils.isNotEmpty(authorizedTaskIds)) {
+            BatchUpdateAlarmParam target = new BatchUpdateAlarmParam();
+            target.setTaskIds(authorizedTaskIds);
+            target.setAlarmSettings(alarm.getAlarmSettings());
+            target.setAlarmRules(alarm.getAlarmRules());
+            target.setEmailReceivers(alarm.getEmailReceivers());
+            alarmService.batchUpdate(target);
```
ply0011, line 1634: **🟡 [correctness] A throw from `batchUpdate` discards the already-computed permission results**

`alarmService.batchUpdate(target)` has no try/catch. If it throws (DB/validation error on any authorized task), the request 500s and the `noAuthMessages` already assembled for permission-denied tasks are lost — no partial-success semantics.

manager/tm-api/src/main/java/com/tapdata/tm/alarm/service/AlarmService.java
```diff
@@ -73,4 +74,6 @@ Page<AlarmListInfoVo> list(String status,
     void updateTaskAlarm(AlarmVO alarm);
 
     void taskRetryAlarm(String taskId,Map<String, Object> params);
+
+    void batchUpdate(BatchUpdateAlarmParam alarm);
```
ply0011, line 78: **🟠 [altitude / cross-repo] New interface method obligates a paired enterprise implementation**

Adding `batchUpdate` to the `AlarmService` interface forces every implementor to provide it. The enterprise `AlarmServiceImpl implements AlarmService` directly and currently has **no** `batchUpdate`, so `tapdata-enterprise/tm-enterprise` will not compile until the paired change lands — and even if it inherited a stub, the enterprise/DAAS feature would be non-functional. Please confirm the enterprise PR ships together with this one.

manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
...
+                responseMessage.setId(taskId);
+                responseMessage.setCode("insufficient.permissions");
+                String msg = String.format("%s.%s", DataPermissionDataTypeEnums.Task, DataPermissionActionEnums.Edit);
+                responseMessage.setMessage(MessageUtil.getMessage("insufficient.permissions", msg, msg));
+                noAuthMessages.add(responseMessage);
+            }
+        }
+
+        if (CollectionUtils.isNotEmpty(authorizedTaskIds)) {
+            BatchUpdateAlarmParam target = new BatchUpdateAlarmParam();
```
ply0011, line 1629: **🟢 [simplification] Redundant full copy of `BatchUpdateAlarmParam`**

A fresh `target` is built copying all four fields just to substitute `authorizedTaskIds`. The incoming `alarm` isn't used afterward, so `alarm.setTaskIds(authorizedTaskIds)` would suffice. As written, any field later added to `BatchUpdateAlarmParam` must be remembered here or it silently drops from the forwarded object.

manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java
```diff
@@ -1581,6 +1584,54 @@ public ResponseMessage<CheckTaskMemoryResult> checkTaskMemoryHeap(@PathVariable(
...
+        List<String> taskIds = alarm.getTaskIds();
+        List<ObjectId> taskObjectIds = taskIds.stream().map(MongoUtils::toObjectId).collect(Collectors.toList());
+        return success(batchUpdateTaskAlarmWithDataPermission(alarm, taskObjectIds, userDetail));
+    }
+
+    private List<MutiResponseMessage> batchUpdateTaskAlarmWithDataPermission(BatchUpdateAlarmParam alarm, List<ObjectId> taskObjectIds, UserDetail userDetail) {
+        List<String> taskIds = alarm.getTaskIds();
+        List<String> authorizedTaskIds = new ArrayList<>();
+        List<MutiResponseMessage> noAuthMessages = new ArrayList<>();
+        for (int i = 0; i < taskIds.size(); i++) {
```
ply0011, line 1604: **🟢 [correctness] Duplicate `taskId`s are processed twice**

The loop iterates `taskIds` without de-duplication, so a repeated id is permission-checked twice and added twice to `authorizedTaskIds`, then forwarded twice to `batchUpdate` — risking double-applied side effects depending on the enterprise impl.

manager/tm-common/src/main/java/com/tapdata/tm/commons/task/dto/alarm/BatchUpdateAlarmParam.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @author <a href="2749984520@qq.com">Gavin'Xiao</a>
+ * @author <a href="https://github.com/11000100111010101100111">Gavin'Xiao</a>
+ * @version v1.0 2026/7/13 17:57 Create
+ * @description
+ */
+@Data
+public class BatchUpdateAlarmParam {
+    List<String> taskIds;
```
ply0011, line 15: **🟢 [conventions] Inconsistent field visibility**

`taskIds` is package-private while the other three fields are `private`. Align the modifier for encapsulation consistency.

```suggestion
    private List<String> taskIds;
```

ply0011 (approved):

PR #3378, opened 2026-07-22
Feat tap 12070 optimize permission functions

weiliang110100:


ply0011:
## Code Review — TAP-12070 优化权限功能

Reviewed at head `961ca88`. 8 findings below (line numbers anchor to the PR head). Severity: 🔴 high / 🟠 medium / 🟡 low.

---

### 🔴 1. `GroupInfoController.tasks()` — NPE when filter has no `syncType` key
`manager/tm/src/main/java/com/tapdata/tm/group/controller/GroupInfoController.java:306`

```java
if(finalFilter.getWhere().get("syncType").equals("sync")){
```
`parseFilter()` never returns null, so the `finalFilter != null` guard is always true and this line always runs. When the `where` map has no `syncType` key (`required=false`, so common — e.g. `filter={"where":{}}`), `get("syncType")` returns `null` and `null.equals("sync")` throws `NullPointerException` → HTTP 500. `getWhere()` itself can also be null (`{"where":null}`). The old code (`if(filter==null) filter=new Filter();` then call the service) worked for any filter.

**Fix:** `"sync".equals(where == null ? null : where.get("syncType"))`.

### 🔴 2. User `DELETE` endpoint checks `Edit` action instead of `Delete`
`manager/tm/src/main/java/com/tapdata/tm/user/controller/UserController.java:553`

```java
return dataPermissionCheckOfId(userDetail, id, DataPermissionActionEnums.Edit, () -> {
    userService.delete(id, userDetail);
```
`DataPermissionDataTypeEnums.User` defines View/Edit/Delete and the init JSON wires `v2_user_management_menu_all_data_Delete`. Because delete authorizes on `Edit`:
- a role granted only **Delete** (but not Edit) is wrongly **denied**;
- a role granted only **Edit** can **delete** users — an over-permission / privilege escalation.

Should be `DataPermissionActionEnums.Delete`. (`updateUserInfo` at :135 correctly uses `Edit`.)

### 🟠 3. Batch stop/delete/renew abort mid-loop on first denial (Task)
`manager/tm/src/main/java/com/tapdata/tm/task/controller/TaskController.java:143`

`dataPermissionCheckOfIds` loops per id; `dataPermissionCheckOfId`'s unauthorized path throws `BizException`. With `taskIds=[A,B]` where the user may act on A but not B: A is stopped/deleted first, then B throws → the whole request returns HTTP 500 **after A was already mutated**, ids after the failure are skipped, and the per-item `MutiResponseMessage` result list is lost. It also calls the batch service **once per id** (N calls / N permission queries) instead of one batched call. Previously a single menu-level check ran once and `batch*` processed all ids together, returning per-item results.

### 🟠 4. Same batch abort defect (shared cache)
`manager/tm/src/main/java/com/tapdata/tm/task/controller/CacheTaskController.java:323`

`checkDataPermissions` has the identical per-id loop — first unauthorized id throws, aborting the batch after earlier ids were mutated and losing per-item results. Fix together with #3.

### 🟠 5. Unauthorized `/tasks` and `/apis` throw a raw `RuntimeException`
`manager/tm/src/main/java/com/tapdata/tm/group/controller/GroupInfoController.java:314` (see `dataPermissionUnAuth()` at :50)

Both endpoints pass the no-arg `this::dataPermissionUnAuth`, which throws `new RuntimeException("Un auth")`, instead of the `BizException("insufficient.permissions", ...)` used everywhere else in this PR. A non-root user lacking the SyncTack/MigrateTack/Modules "all data" menu permission gets an unhandled HTTP 500 with `Un auth` rather than the standard localized error. Leftover duplicate of the 3-arg `dataPermissionUnAuth`.

### 🟡 6. Non-cloud `findById`/`findOne` drop the `userDetail` owner filter — verify intent
`manager/tm/src/main/java/com/tapdata/tm/externalStorage/service/ExternalStorageServiceImpl.java:245` (also :254, :268)

For non-cloud, these now call `super.findById(objectId, field)` / `super.findOne(query)` **without** `userDetail`, removing the per-user data filter so any user can resolve any external-storage record by id/query (password masked, but name/type/uri exposed). If external storage is intentionally shared on-prem this is the intended fix — but it silently changes the authorization boundary across three overloads and should be confirmed.

### 🟡 7. `updateShareCacheTask` no longer restarts the task after edit — verify intent
`manager/tm/src/main/java/com/tapdata/tm/task/service/TaskServiceImpl.java:824`

The `start(taskDto.getId(), user)` call was removed, so editing a shared-cache task now only persists the change without (re)starting it to apply the new config at runtime. The added test asserts `never().start()`, so this looks intentional, but it's a runtime behavior change bundled into a permissions PR — worth confirming it matches product intent.

### 🟡 8. `checkInspect` runs one permission query per action
`manager/tm/src/main/java/com/tapdata/tm/inspect/controller/InspectResultController.java:96`

`checkInspect` now calls `DataPermissionHelper.check` once per action with a non-null `id`, and each call triggers a separate `setDataActions` lookup — N DB round-trips for N actions, versus the previous single `findDataActions` call followed by in-memory `contains` checks. Consolidate to one lookup.

---

*Findings #6 and #7 are deliberately-tested changes that are likely intended — flagged so the author confirms the authorization / runtime-behavior shift is wanted.*

ply0011 (changes requested):
需要修改。详见上方 code review 评论（https://github.com/tapdata/tapdata/pull/3378#issuecomment-5089535790）。

阻塞项（须修复）：
- 🔴 #1 GroupInfoController.tasks():306 — filter 无 syncType 时 `null.equals("sync")` NPE，接口 500。
- 🔴 #2 UserController.delete():553 — 删除用户却校验 Edit 权限（应为 Delete），存在越权/权限不符。
- 🟠 #3 / #4 批量 stop/delete/renew（TaskController:143 与 CacheTaskController:323）— 逐个鉴权在首个无权 id 处抛异常中断，前面的 id 已产生副作用且丢失 per-item 结果。

请确认项：
- 🟡 #6 ExternalStorageServiceImpl 非 cloud 去掉 userDetail 过滤（授权边界变化）。
- 🟡 #7 updateShareCacheTask 编辑后不再 start（运行时行为变化）。

weiliang110100:
#1 – GroupInfoController NPE
Added a null-safe syncType check. Missing where or syncType now defaults to migration-task permissions instead of returning HTTP 500.

#2 – Incorrect user delete permission
Changed user deletion authorization from Edit to Delete, preventing both false denials and privilege escalation.

#3 / #4 – Batch operations aborted on permission denial
Unauthorized IDs now produce per-item insufficient.permissions results without interrupting subsequent items. Authorized operations execute within their permission context, while service-level business exceptions retain their original behavior.

#7 – Shared-cache restart after editing

ply0011 (approved):

PR #3407, opened 2026-08-04
Fix tap 12225 develop

weiliang110100:


manager/tm/src/main/java/com/tapdata/tm/config/ObsMongoConfig.java
```diff
@@ -57,33 +65,82 @@ public CompletableFuture<MongoTemplate> mongoTemplate() throws Exception {
...
-                            new BsonDocument("tags.nodeId", new BsonInt32(1))));
-
-                    agentMeasurementV2.createIndex(Indexes.compoundIndex(new BsonDocument("grnty", new BsonInt32(1)),
-                            new BsonDocument("tags.engineId", new BsonInt32(1)),
-                            new BsonDocument("tags.type", new BsonInt32(1))));
-
-                    agentMeasurementV2.createIndex(new BsonDocument("date", new BsonInt32(-1)), new IndexOptions().expireAfter(7L, TimeUnit.DAYS));
-                }
+                MongoCollection<Document> agentMeasurementV2 = mongoTemplate.getCollection(MeasurementEntity.COLLECTION_NAME);
+                initializeAgentMeasurementIndexes(agentMeasurementV2);
```
ply0011: **High** — this call has no error isolation, and a failure here now blocks TM startup.

The four legacy `createIndex` calls inside `initializeAgentMeasurementIndexes` are still gated on `estimatedDocumentCount() == 0`, but the `listIndexes()` + two `ensureIndex` calls below them are **unconditional** — they run on every startup, including on populated production collections where this code path was previously unreachable.

Any exception out of them propagates into `catch (Exception e) { throw new RuntimeException(e); }` on line 72, which completes the `obsMongoTemplate` future exceptionally. `MeasurementServiceV2Impl`'s constructor blocks on `mongoTemplateCompletableFuture.get()` (`MeasurementServiceV2Impl.java:121`), so the bean fails to construct and **TM does not start**.

Realistic triggers: an existing index with the same keys under a different name (`IndexOptionsConflict`, code 85 — see the `sameIndexKeys` comment below), an obs DB user without `createIndex` privilege, or an interrupted index build.

A missing index should degrade dashboard performance, not block boot. Suggest try/catch + `logger.error` around the `ensureIndex` calls.

manager/tm/src/main/java/com/tapdata/tm/config/ObsMongoConfig.java
```diff
@@ -57,33 +65,82 @@ public CompletableFuture<MongoTemplate> mongoTemplate() throws Exception {
...
+                return;
+            }
+            if (sameIndexKeys(existingKeys, indexKeys)) {
+                logger.info("Index {} already exists in collection {} as {}, skipping creation",
+                        indexName, MeasurementEntity.COLLECTION_NAME, existingIndex.getString("name"));
+                return;
+            }
+        }
+
+        collection.createIndex(indexKeys, new IndexOptions().name(indexName).background(true));
```
ply0011: **Medium** — `background(true)` has been a no-op since MongoDB 4.2, so this blocks startup for the full build.

`createIndex` is synchronous and the driver waits for the build to complete regardless of the `background` flag (the option is accepted and ignored on 4.2+). On the large `AgentMeasurementV2` collections this PR is optimizing for, the first startup after upgrade builds two compound indexes inline, while `MeasurementServiceV2Impl`'s constructor sits on an untimed `mongoTemplateCompletableFuture.get()`. TM startup hangs for the entire build — minutes or more on a big obs DB.

Consider moving index creation off the startup path (e.g. an async post-start task), or at minimum a bounded wait plus a degraded-mode path.

manager/tm/src/main/java/com/tapdata/tm/config/ObsMongoConfig.java
```diff
@@ -57,33 +65,82 @@ public CompletableFuture<MongoTemplate> mongoTemplate() throws Exception {
...
+        existingIndexes.add(new Document("name", indexName).append("key", indexKeys));
+        logger.info("Created index {} in collection {}", indexName, MeasurementEntity.COLLECTION_NAME);
+    }
+
+    private boolean sameIndexKeys(Document left, Document right) {
+        return left != null && right != null
+                && new ArrayList<>(left.entrySet()).equals(new ArrayList<>(right.entrySet()));
```
ply0011: **Medium** — this comparison is type-strict, which defeats the "same keys under a different name" guard in `ensureIndex`.

`Map.Entry.equals` compares values with `equals`, so `Integer(1).equals(Double(1.0))` is `false`. The BSON type of an index key value depends on how the index was created: `indexKeys` here builds `Integer 1/-1`, but `listIndexes()` returns `Double 1.0/-1.0` for indexes created by the legacy `mongo` shell or restored via `mongodump`/`mongorestore`.

On those clusters this returns `false`, `ensureIndex` falls through to `collection.createIndex(...)` on line 137, and MongoDB — whose own key-pattern comparison *is* numeric — rejects it as "same key pattern, different name" with `IndexOptionsConflict`. Combined with the missing error isolation at line 69, that becomes a TM startup failure on exactly the older clusters.

Compare numerically, e.g. normalize each value through `Number#doubleValue()` before comparing.

manager/tm/src/main/java/com/tapdata/tm/monitor/service/MeasurementServiceV2Impl.java
```diff
@@ -1454,30 +1458,45 @@ public MeasurementEntity findLastMinuteByTaskId(String taskId) {
...
-                    .flatMap(samples -> samples.stream()
-                            .filter(Objects::nonNull)
-                            .filter(sample -> sample.getDate() != null)
-                            .max(Comparator.comparing(Sample::getDate)))
-                    .ifPresent(latest -> result.put(tid, latest));
-        }
+        forEachTaskIdBatch(normalizedTaskIds, taskIdBatch -> {
+            Criteria criteria = Criteria.where(MetricCons.F_GRANULARITY).is(Granularity.GRANULARITY_MINUTE)
+                    .and(PATH_TAGS_TYPE).is(SAMPLE_TYPE_TASK)
+                    .and(PATH_TAGS_TASK_ID).in(taskIdBatch);
```
ply0011: **High** — no date lower bound: this reads the entire 7-day TTL window for every task in the batch.

The criteria filters only on granularity / type / taskId. The pipeline below then does `$sort` + `$group ... .first("$$ROOT")`, and before MongoDB 8.0 there is no `DISTINCT_SCAN` optimization for `$group`/`$first` — so the server must stream and FETCH every matching minute-granularity document rather than stopping at one per group.

Minute docs are one per task per minute (up to 200 samples each), ≈10k docs per task per 7 days. A dashboard refresh with 100 running tasks reads on the order of 1M documents.

Note the irony: `idx_taskId_grnty_type_date`, added by this same PR, would let the *previous* per-task `findLastMinuteByTaskId` do exactly one index seek + `limit 1` per task. Adding `.and(MetricCons.F_DATE).gte(<now - N minutes>)` here recovers most of that.

manager/tm/src/main/java/com/tapdata/tm/monitor/service/MeasurementServiceV2Impl.java
```diff
@@ -1454,30 +1458,45 @@ public MeasurementEntity findLastMinuteByTaskId(String taskId) {
...
+            Aggregation aggregation = Aggregation.newAggregation(
+                    Aggregation.match(criteria),
+                    Aggregation.sort(Sort.by(Sort.Order.asc(PATH_TAGS_TASK_ID), Sort.Order.desc(MetricCons.F_DATE))),
+                    Aggregation.group(PATH_TAGS_TASK_ID).first("$$ROOT").as("latestRecord"),
+                    Aggregation.replaceRoot().withValueOf("$latestRecord"),
+                    Aggregation.project(MetricCons.F_TAGS, MetricCons.F_SAMPLES));
```
ply0011: **Medium** — no `allowDiskUse`, so an unusable index turns a slow endpoint into a failing one.

If the `$sort {tags.taskId: 1, date: -1}` on line 1472 can't use `idx_taskId_grnty_type_date` — silently skipped by `ensureIndex`'s same-name/different-keys branch, still building, or never created due to permissions — it degrades to a blocking in-memory sort, and MongoDB aborts past the 100MB limit with `QueryExceededMemoryLimitNoDiskUseAllowed` (292). The dashboard endpoint then returns 500. The previous per-task implementation was only slow, never fatal.

```java
Aggregation aggregation = Aggregation.newAggregation(
        // ... stages ...
).withOptions(AggregationOptions.builder().allowDiskUse(true).build());
```

Also worth logging a warning (not just `info`) from `ensureIndex` when it skips creation, so this failure mode is diagnosable from the TM log.

ply0011 (commented):
Inline version of the review I posted earlier as a plain comment — same five findings, now anchored to the specific lines.

Two `ObsMongoConfig` issues (unguarded `ensureIndex` on the startup path + type-strict `sameIndexKeys`) compound into a **TM startup failure** on clusters whose `AgentMeasurementV2` indexes were created by the legacy `mongo` shell or a `mongorestore`. The `MeasurementServiceV2Impl` aggregation is correct but unbounded in time, so it can read ~1M docs per dashboard refresh — while the index this PR adds would have made the *old* per-task lookup a single seek.

The pipeline itself renders correctly under Spring Data MongoDB 5.0.x (`first("$$ROOT")`, `replaceRoot().withValueOf("$latestRecord")`, `group("tags.taskId")`, and the post-`$replaceRoot` `project`), and `normalizeTaskIds` / `forEachTaskIdBatch` are semantically equivalent to the previous batching behavior. No issues in the test changes.

ply0011 (changes requested):
Requesting changes on the strength of the two High findings — details are in the inline comments.

### High

- **`ObsMongoConfig.java:69`** — the `listIndexes()` + two `ensureIndex` calls are unconditional (unlike the four legacy `createIndex` calls, which remain gated on `estimatedDocumentCount() == 0`) and have no error isolation. Anything they throw becomes `RuntimeException` → the `obsMongoTemplate` future completes exceptionally → `MeasurementServiceV2Impl`'s constructor blocks on `.get()` and fails → **TM does not start**. A missing index should cost dashboard performance, not boot.
- **`MeasurementServiceV2Impl.java:1467`** — the new aggregation has no date lower bound, so `$sort` + `$group/$first` must FETCH the entire 7-day TTL window for every task in the batch (no `DISTINCT_SCAN` before MongoDB 8.0). ~1M documents per dashboard refresh at 100 running tasks — while `idx_taskId_grnty_type_date`, added by this same PR, would have made the *previous* per-task lookup a single index seek.

### Medium

- **`ObsMongoConfig.java:142-144`** — `sameIndexKeys` compares entries with `Map.Entry.equals`, so `Integer 1` ≠ `Double 1.0`. Indexes created by the legacy `mongo` shell or a `mongorestore` slip past the "same keys, different name" guard into `createIndex`, which MongoDB then rejects with `IndexOptionsConflict`. Compounds with the first High into a startup failure on exactly those clusters.
- **`ObsMongoConfig.java:137`** — `background(true)` is ignored on MongoDB ≥ 4.2, so the first startup after upgrade blocks on the full index build against an untimed `.get()`.
- **`MeasurementServiceV2Impl.java:1470`** — no `allowDiskUse`; if the sort can't use the index it becomes a blocking in-memory sort and fails with `QueryExceededMemoryLimitNoDiskUseAllowed` (292), turning a previously-slow endpoint into a 500.

### Not blocking

The pipeline renders correctly under Spring Data MongoDB 5.0.x, `normalizeTaskIds` / `forEachTaskIdBatch` are semantically equivalent to the previous batching, and the test changes look fine. The direction of this PR is right — the two High items are about failure modes it introduces on the way there.

weiliang110100:
## Code Review 修复说明

已在提交 [`7a728677`](https://github.com/tapdata/tapdata/commit/7a728677b4ad569a1800d3ac5a3192933e2712d1) 中完成本轮 review 提出的修复。

### 修复内容

- [x] **High：索引初始化异常导致 TM 启动失败**
  - 将 Dashboard 索引初始化从 `obsMongoTemplate` 启动链路中拆出，改为异步执行，索引构建不再阻塞 TM 启动。
  - 分别隔离 `listIndexes` 和每个 `createIndex` 的异常；单个索引初始化失败只记录日志，不再导致 MongoTemplate future 异常完成。

- [x] **High：最新监控数据聚合扫描整个 7 天 TTL 窗口**
  - 为运行中任务的分钟级监控查询增加 `[now - 5 min, now]` 时间范围。
  - 所有 taskId 批次复用同一组起止时间，避免批次间窗口漂移。
  - 保留每批 100 个 taskId 的聚合方式，使新增复合索引能够将扫描范围限制在最近 5 分钟。

- [x] **Medium：历史索引方向值的数值类型不一致**
  - `sameIndexKeys` 对 `Integer`、`Long`、`Double` 等 `Number` 类型按数值比较，兼容 legacy mongo shell 和 `mongorestore` 创建的索引，避免误触发 `IndexOptionsConflict`。

- [x] **Medium：`background(true)` 无效且可能造成启动等待**
  - 移除 `background(true)`；索引构建已移到异步任务，不再阻塞应用启动。

- [x] **Medium：聚合排序未允许磁盘使用**
  - 聚合增加 `allowDiskUse(true)`，避免索引未命中时因内存排序上限导致接口返回 500。

### 验证结果

```bash
mvn -pl tm -am \
  -Dtest=ObsMongoConfigTest,MeasurementServiceV2ImplTest \
  -Dsurefire.failIfNoSpecifiedTests=false test
```

- Tests run: **76**
- Failures: **0**
- Errors: **0**
- Result: **BUILD SUCCESS**

ply0011 (approved):

PR #3423, opened 2026-08-12
TAP-12157:Fix data loss issue during shared mining

weiliang110100:


iengine/api/src/main/java/io/tapdata/flow/engine/V2/sharecdc/ShareCDCOffset.java
```diff
@@ -15,12 +16,12 @@ public class ShareCDCOffset implements Serializable {
 	private Object streamOffset;
 
 	public ShareCDCOffset(Map<String, Long> sequenceMap, Object streamOffset) {
-		this.sequenceMap = sequenceMap;
+		this.sequenceMap = null == sequenceMap ? null : new HashMap<>(sequenceMap);
```
ply0011, line 19: **[medium] The snapshot still races across ReadRunner threads — data loss window remains.**

The copy fixes the same-table case, but `sequenceMap` in `ShareCdcPDKTaskReader` is shared by *all* `ReadRunner` threads, while `tapEventWrapper()` snapshots the whole map.

In the read loop (`ShareCdcPDKTaskReader#read`, ~line 513):
```java
sequenceMap.put(tableName, iterator.getSequence()); // getSequence() == next unread sequence
documents.forEach(doc -> enqueue(tapEventWrapper(doc)));
```
Thread B does `put(tableB, seqB)` and is preempted before `enqueue(eventB)`. Thread A wraps `eventA` in that window, so `eventA` carries `{tableB: seqB}` even though `eventB` has not entered the queue yet. Downstream processes `eventA`, persists that offset, and the task dies before `eventB` is processed. On resume `ConstructRingBuffer.find({sequence: seqB})` starts *at* seqB (`RingBufferIterator` increments after reading, so the stored value is the next unread sequence) — `eventB` is skipped. Same data loss the PR is fixing, just a narrower window.

Swapping the two lines (enqueue first, then `sequenceMap.put`) turns the residual race from loss into replay, which is safe under at-least-once.

ply0011, line 19: **[low] O(#tables) copy on the per-event hot path, and each in-flight event now retains its own copy.**

`new ShareCDCOffset(sequenceMap, offsetObj)` is called from `ShareCdcPDKTaskReader#tapEventWrapper` for *every* CDC record, so this constructor copy is now O(number of tables mined by the task) per record — previously O(1) (shared reference). A log-collector task mining ~1000 tables at ~10k events/s copies ~10M map entries per second.

Memory too: before, all queued `ShareCDCReaderEvent`s shared one map; now every event in the reader queue and in every downstream buffer holds a full `HashMap` of all tables' sequences (1000 events in flight x 1000 tables = 1M retained entries).

If that turns out to matter, keep a single immutable snapshot in the reader that is rebuilt when a sequence changes and shared by the events created from it, instead of copying inside the offset constructor.

iengine/api/src/main/java/io/tapdata/flow/engine/V2/sharecdc/ShareCDCOffset.java
```diff
@@ -15,12 +16,12 @@ public class ShareCDCOffset implements Serializable {
...
 		this.streamOffset = streamOffset;
 	}
 
 	public Map<String, Long> getSequenceMap() {
-		return sequenceMap;
+		return null == sequenceMap ? null : new HashMap<>(sequenceMap);
```
ply0011, line 24: **[low] The getter copy is redundant now that the constructor copies.**

`sequenceMap` is private, has no setter, and is never mutated after construction, so the field is already a snapshot nobody else holds a reference to. Copying again on every read only adds allocation, and it silently makes any future `offset.getSequenceMap().put(...)` a no-op instead of a compile-time signal. `Collections.unmodifiableMap` (or just returning the field) says the same thing without the copy.

Note the null pass-through is preserved on both sides, so `ShareCdcPDKTaskReader:114` (`generateSequenceMap(((ShareCDCOffset) offset).getSequenceMap())`) still NPEs on a null map — pre-existing, but worth a null guard there if a null-sequenceMap offset can ever be persisted.

ply0011 (changes requested):
Reviewed the defensive-copy fix. The direction is right — snapshotting the sequence map per event does close the main loss path where every emitted offset pointed at the newest sequences — but one issue keeps the fix incomplete:

- **medium** `ShareCDCOffset:19` — `sequenceMap` is shared by all `ReadRunner` threads, and `tapEventWrapper` snapshots the whole map. A thread that has already done `sequenceMap.put(table, seq)` but not yet `enqueue(...)` leaks its advanced sequence into another thread's event, so persisting that event's offset can still skip an unenqueued record on resume. Swapping `put` and `enqueue` in `ShareCdcPDKTaskReader#read` makes the residual race a replay instead of a loss.
- **low** `ShareCDCOffset:19` — the copy is now O(#tables) per CDC record, and every in-flight event retains its own full map; worth a shared immutable snapshot for wide mining tasks.
- **low** `ShareCDCOffset:24` — the getter copy is redundant (field is private, never mutated post-construction); `unmodifiableMap` conveys the same guarantee without the allocation.

PR #3473, opened 2026-09-07
Fix English task-check output to localize default node names

weiliang110100:
Constraint: Preserve user-defined node names while localizing only NodeEnum defaults.

Rejected: Per-strategy ad hoc translations would keep output paths inconsistent.

Confidence: medium

Scope-risk: moderate

Tested: git diff --check; Maven target test attempted but baseline missing-class compilation errors blocked execution.

Not-tested: MessageUtilDagCheckLocalizationTest runtime assertions.


HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **0.0**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix/TAP-12611&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **6.6**
- Status: **ERROR**, MetricKey: **new_duplicated_lines_density**, ActualValue: **19.69697**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix/TAP-12611&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_coverage**, ActualValue: **19.7**
- Status: **ERROR**, MetricKey: **new_duplicated_lines_density**, ActualValue: **19.69697**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix/TAP-12611&id=tapdata

HarsenLin:

- Status: **ERROR**, MetricKey: **new_duplicated_lines_density**, ActualValue: **19.69697**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=fix/TAP-12611&id=tapdata

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/dagcheckstrategy/UnionNodeStrategyImpl.java
```diff
@@ -45,23 +43,11 @@ public List<TaskDagCheckLog> getLogs(TaskDto taskDto, UserDetail userDetail, Loc
...
 
                     if (CollectionUtils.isEmpty(result) || result.stream().anyMatch(log -> nodeId.equals(log.getNodeId()))) {
-                        TaskDagCheckLog log = TaskDagCheckLog.builder().taskId(taskId).checkType(templateEnum.name())
-                                .grade(Level.INFO).nodeId(nodeId)
-                                .log(MessageFormat.format(MessageUtil.getDagCheckMsg(locale, "UNION_PASS"), name))
-                                .build();
-                        log.setCreateAt(now);
-                        log.setCreateUser(userId);
-                        result.add(log);
+                        result.add(DagCheckLogs.of(taskId, nodeId, userId, now, Level.INFO, templateEnum, locale, "UNION_PASS", name));
```
ply0011: **Missing `UNION_PASS` in `dagCheck_zh_CN.properties` now degrades differently (and can NPE).**

This rewrite changes `MessageFormat.format(getDagCheckMsg(locale, "UNION_PASS"), name)` into `getDagCheckMsg(locale, "UNION_PASS", name)`. Those are not equivalent when the key is missing from the bundle — and `UNION_PASS` *is* missing from `manager/tm/src/main/resources/dagCheck_zh_CN.properties` (it exists only in `en_US` and `zh_TW`).

`MessageUtil.getString` takes the `msg == null` branch:
```java
if (msg == null) {
    if (params != null && params.length > 0) {
        return Streamable.of(params).stream().map(Object::toString).collect(joining(","));
    }
    return resourceId;
}
```

So for a zh_CN user with a Union node:
* before: the log line read `UNION_PASS` (broken, but self-identifying),
* now: the log line is just the raw node name, e.g. `Union节点`.

Worse, the two branches above interact: when `name` is `null`, the `StringUtils.isEmpty(name)` branch fires *and* this branch fires (the `anyMatch(nodeId…)` predicate is true), so `params = {null}` reaches `map(Object::toString)` → **NPE**, which propagates out of `dagCheck` and 500s `GET /api/task-console`. Previously `MessageFormat.format("UNION_PASS", (Object) null)` was harmless.

Simplest fix is to add `UNION_PASS` to `dagCheck_zh_CN.properties`; hardening `getString`'s `msg == null` path against null params would also be worthwhile since this PR now routes every strategy through it.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
```diff
@@ -115,7 +127,8 @@ public TaskDagCheckLogVo getLogs(TaskLogDto dto, UserDetail userDetail, Locale l
                 .filter(n -> (n instanceof MigrateJsProcessorNode || n instanceof JsProcessorNode || n instanceof PyProcessNode || n instanceof MigratePyProcessNode) && !(n instanceof CustomProcessorNode))
                 .findFirst();
         if (jsNode.isPresent()) {
-            List<TaskDagCheckLog> jsNodeLog = monitoringLogsService.getJsNodeLog(taskDto.getTransformTaskId(), taskDto.getName(), NodeEnum.valueOf(jsNode.get().getType()).getNodeName());
+            List<TaskDagCheckLog> jsNodeLog = monitoringLogsService.getJsNodeLog(taskDto.getTransformTaskId(), taskDto.getName(),
+                    MessageUtil.localizeDagNodeName(locale, NodeEnum.valueOf(jsNode.get().getType()).getNodeName()));
```
ply0011: **The localized name is injected into a hard-coded Chinese wrapper, producing mixed output.**

`MonitoringLogsServiceImpl.getJsNodeLog` formats with a literal template that already appends 节点:

```java
String message = "{0}【{1}】【{2}节点】{3}";
info.setLog(MessageFormat.format(message, date, taskName, nodeName, …));
```

With the old Chinese `增强JS节点` this rendered `【增强JS节点节点】` (already odd). After this change, an English-locale request renders `【Enhanced JS node节点】` — English name glued to a Chinese suffix inside Chinese brackets, in exactly the scenario this PR is trying to fix (English UI, task with a JS/Python node).

If the goal is English output here, `getJsNodeLog`'s template needs to be moved into the `dagCheck` bundle too; otherwise it's better to keep passing the untranslated `NodeEnum` name so at least the line stays internally consistent.

manager/tm-api/src/main/java/com/tapdata/tm/utils/MessageUtil.java
```diff
@@ -107,7 +108,29 @@ private static String getString(Locale locale, String resourceId, String msg, Ob
 
 	public static String getDagCheckMsg(Locale locale, String resourceId, Object... params){
 		String msg = getStringOrNull(getResourceBundle(locale, "dagCheck"), resourceId);
-		return getString(locale, resourceId, msg, params);
+		return getString(locale, resourceId, msg, localizeDagCheckParams(locale, params));
+	}
+
+	public static String localizeDagNodeName(Locale locale, String nodeName) {
```
ply0011: **Content-based matching cannot tell a default node name from user data — it rewrites both.**

`localizeDagCheckParams` runs *every* `String` parameter through this method, and the match is on the literal Chinese text. But the parameters reaching `getDagCheckMsg`/`createLog` are not only node names: they are also connection names (`SOURCE_CONNECT_INFO` → `connectionDto.getName()`), table names (`TABLE_EDIT_NODE_INFO` → `current`), field names (`TARGET_SETTING_CHECK_FIELD` → `k`), and JSON blobs.

Concrete failure: this product's users routinely have Chinese identifiers. A user with a MySQL table literally named `表节点`, or who deliberately renamed a node to `数据源节点`, gets that string silently replaced with `Table node` / `Data source node` in the English check output — the PR description's stated constraint ("Preserve user-defined node names") cannot be enforced by string equality.

The robust version is to translate at the source, where the type is known — i.e. where the default name is produced from `NodeEnum` (`JsSettingStrategyImpl`'s `NodeEnum.valueOf(node.getType()).getNodeName()`, and `getLogs`' node map, where `node.getType()` is available) — rather than pattern-matching finished parameter strings.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
```diff
@@ -54,6 +54,8 @@
 @Setter(onMethod_ = {@Autowired})
 public class TaskDagCheckLogServiceImpl implements TaskDagCheckLogService {
 
+    private final ThreadLocal<Locale> dagCheckLocale = new ThreadLocal<>();
```
ply0011: **This ThreadLocal doesn't cover the call sites in this very file, so one response can mix locales.**

`dagCheck` sets and restores it around the strategy loop, but `getLogs` calls `createLog` three more times (`MODEL_PROCESS_FAIL` ×2, `MODEL_PROCESS_INFO`) *after* `dagCheck` has already cleared it — see lines 150/160/177. Those fall through to `MessageUtil.getLocale()`, which reads `RequestContextHolder` and hard-defaults to `en_US` when there is no request. So the strategy logs are localized with the `locale` argument while the model-process logs in the same `TaskDagCheckLogVo` are localized with whatever the request happens to say (or `en_US`).

Every one of those call sites has `locale` in scope. Adding a `Locale` parameter to `createLog` (or an overload) removes the ThreadLocal, the save/restore dance, and the inconsistency in one smaller diff.

manager/tm/src/main/resources/dagCheck_en_US.properties
```diff
@@ -77,4 +77,38 @@ TARGET_SETTING_SELECT_FIELD=$date [Source Type Mapping Detection] [{0}] [{1}] [{
...
-MIGRATE_UNION_PASS=$date [$taskName][Union For Multiple Table]: Multiple table union has passed the check.
\ No newline at end of file
+MIGRATE_UNION_PASS=$date [$taskName][Union For Multiple Table]: Multiple table union has passed the check.
+NODE_NAME_mem_cache=Shared cache node
+NODE_NAME_database=Data source node
+NODE_NAME_table=Table node
+NODE_NAME_hazelcastIMDG=Shared mining assignment node
+NODE_NAME_logCollector=Shared mining node
+NODE_NAME_aggregation_processor=Aggregation node
+NODE_NAME_cache_lookup_processor=Cache lookup processor node
```
ply0011: **Dead key: `NodeEnum.cache_lookup_processor`'s node name is the empty string.**

```java
cache_lookup_processor(""),
```

`localizeDagNodeName` returns early on `StringUtils.isBlank(nodeName)`, so no input can ever reach the `NODE_NAME_cache_lookup_processor` lookup. Harmless, but it hides the fact that cache-lookup nodes have no default name to localize in the first place — worth either dropping this line or fixing the enum.

ply0011 (changes requested):
Careful pass over the whole diff. The mechanical refactor into `DagCheckLogs` is faithful — I checked argument order, grade, nodeId and createAt on all 14 converted strategies and they match the code they replace, and all 34 `NODE_NAME_*` keys line up with `NodeEnum` constants (the duplicated Chinese names all map to the same English string, so the first-match-wins loop is safe).

Two things need fixing before merge:

**Blocking**
1. `UnionNodeStrategyImpl:50` — `MessageFormat.format(getDagCheckMsg(k), args)` → `getDagCheckMsg(k, args)` is not equivalent when the key is missing, and `UNION_PASS` is genuinely absent from `dagCheck_zh_CN.properties`. zh_CN output degrades from `UNION_PASS` to a bare node name, and a null node name reaches `map(Object::toString)` in `MessageUtil.getString` → NPE out of `dagCheck` → 500 on `GET /api/task-console`.
2. `MessageUtil.localizeDagNodeName` — matching on literal Chinese text cannot distinguish a `NodeEnum` default from user data. Connection names, table names and field names all flow through `localizeDagCheckParams`, so a table named `表节点` or a node the user deliberately named `数据源节点` gets silently rewritten in English output, which is the opposite of the PR's stated constraint. Translating where `node.getType()` is known would be exact.

**Non-blocking**
3. `TaskDagCheckLogServiceImpl:131` — the localized name is passed into `getJsNodeLog`'s hard-coded `{0}【{1}】【{2}节点】{3}`, yielding `【Enhanced JS node节点】` in English mode.
4. `TaskDagCheckLogServiceImpl:57` — the `ThreadLocal` doesn't reach the three `createLog` calls in `getLogs` itself (lines 150/160/177 run after `dagCheck` restored it), so those silently fall back to `en_US`. `locale` is in scope at every call site; a parameter would be smaller and correct.
5. `dagCheck_en_US.properties:87` — `NODE_NAME_cache_lookup_processor` is unreachable (`NodeEnum.cache_lookup_processor` has an empty node name).

I could not build to verify: `tm-common` cannot resolve `io.tapdata:tapdata-api:2.0.10-SNAPSHOT` locally, the same baseline failure noted in the PR description. Worth getting CI green on the new tests before merging.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/dagcheckstrategy/UnionNodeStrategyImpl.java
weiliang110100: Added `UNION_PASS` to `dagCheck_zh_CN.properties`. `getDagCheckMsg` now returns the resource id when the key is missing instead of joining params, and `getString`'s `msg == null` path uses `Objects.toString` so a null param no longer NPEs out of `dagCheck`.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
weiliang110100: Stopped passing a localized name into `getJsNodeLog`. It still uses the hard-coded `{2}节点` wrapper, so the untranslated `NodeEnum` default keeps that line internally consistent.

manager/tm-api/src/main/java/com/tapdata/tm/utils/MessageUtil.java
weiliang110100: `localizeDagNodeName` now takes `nodeType` and only translates when the name equals that type's `NodeEnum` default. Call sites pass a `DagNodeName`; plain strings (connection / table / field names) are left unchanged, so a table named `表节点` or a user-renamed `数据源节点` is not rewritten.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
weiliang110100: Removed the `ThreadLocal`. `createLog` now takes `Locale`, and the `MODEL_PROCESS_*` calls in `getLogs` pass the same `locale` argument as the strategy loop.

manager/tm/src/main/resources/dagCheck_en_US.properties
weiliang110100: Dropped `NODE_NAME_cache_lookup_processor`; `NodeEnum.cache_lookup_processor` has an empty name so the key was unreachable.

weiliang110100:
review 意见已在 `9fa5a15771` 处理。

**Blocking**
1. `UNION_PASS` 缺失导致 zh_CN 输出退化、空节点名 NPE：已补 `dagCheck_zh_CN.properties`。`getDagCheckMsg` 在 key 缺失时返回 resource id，不再拼接 params；`getString` 的 `msg == null` 路径用 `Objects.toString`，null 参数不再从 `dagCheck` 抛 NPE。
2. 按中文文本匹配会改写用户数据：`localizeDagNodeName` 改为接收 `nodeType`，仅当名称等于该类型 `NodeEnum` 默认名才翻译。调用方传 `DagNodeName`；连接名 / 表名 / 字段名等普通字符串保持原样。

**Non-blocking**
3. `getJsNodeLog` 硬编码 `{2}节点` 导致中英混排：不再传入已翻译名称，继续用未翻译的 `NodeEnum` 默认名。
4. `ThreadLocal` 覆盖不到 `getLogs` 里的 `MODEL_PROCESS_*`：已删除 ThreadLocal，`createLog` 增加 `Locale` 参数，与 strategy 循环使用同一个 locale。
5. `NODE_NAME_cache_lookup_processor` 不可达：已删除。`NodeEnum.cache_lookup_processor` 默认名为空字符串。

**额外**
`DATA_INSPECT_NEED_CLOSED_*` 模板仍是 `%s`，不能走 `MessageFormat`。改回 `String.format`，并在源头按节点类型做默认名本地化。

**测试**
- `tm-api` `MessageUtilDagCheckLocalizationTest`：16 passed
- `tm` localization / strategy tests：38 passed

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
weiliang110100: Follow-up for TAP-12861: `getJsNodeLog` now takes `Locale` and formats `JS_NODE_MONITOR_LOG` from the dagCheck bundle (`{0} [{1}] [{2}] {3}` / `{0}【{1}】【{2}】{3}`), without appending an extra `节点`. The caller passes the node's stored name through `localizeDagNodeName`, so English save-check no longer contains `增强JS节点` or `【Enhanced JS node节点】`.

weiliang110100:
补充：`82.12861` 失败来自 `getJsNodeLog` 写死的 `{0}【{1}】【{2}节点】{3}`，英文 save-check 会混入 `【增强JS节点节点】`。

已在 `e7acaf3ad0` 处理：
- 模板迁入 `dagCheck`（en_US / zh_CN / zh_TW）
- `getJsNodeLog` 增加 `Locale`
- 传入节点实际名称并走 `localizeDagNodeName`（默认名英文化，自定义名保留）

单测：`MonitoringLogsServiceImplTest$GetJsNodeLogTest` + `TaskDagCheckLogServiceImplLocalizationTest` passed。

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/dagcheckstrategy/UnionNodeStrategyImpl.java
ply0011: Verified at e7acaf3a.

- `UNION_PASS` is now present in `dagCheck_zh_CN.properties`.
- `getDagCheckMsg` returns the resource id when the key is missing, so the `msg == null` param-joining branch is no longer reachable from dag-check at all; `getString` additionally uses `Objects.toString(param, "null")`.
- The NPE case (Union node with a null name under zh_CN) is pinned by `MessageUtilDagCheckLocalizationTest#formatsChineseUnionPassAndDoesNotThrowOnNullName`.

Ran locally: tm-api 16 tests, tm 44 tests, all green. Resolving.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
ply0011: Verified at e7acaf3a (follow-up commit included).

The hard-coded `{0}【{1}】【{2}节点】{3}` is gone; `getJsNodeLog` takes a `Locale` and formats `JS_NODE_MONITOR_LOG` from the dagCheck bundle (`{0} [{1}] [{2}] {3}` en_US / `{0}【{1}】【{2}】{3}` zh_CN, zh_TW), with no extra `节点` appended. The caller passes the node stored name through `localizeDagNodeName`, so an English request no longer yields `【Enhanced JS node节点】` and a Chinese one no longer yields `【增强JS节点节点】`.

`MonitoringLogsServiceImplTest$GetJsNodeLogTest` asserts both directions and is green locally. Resolving.

manager/tm-api/src/main/java/com/tapdata/tm/utils/MessageUtil.java
ply0011: Verified at e7acaf3a.

`localizeDagCheckParams` now rewrites only `DagNodeName` params, and `localizeDagNodeName` translates only when the name equals that node type own `NodeEnum` default. Connection names, table names, field names and JSON blobs reach `createLog`/`getDagCheckMsg` as plain Strings and pass through untouched — a table literally named `表节点` stays `表节点` (`MessageUtilDagCheckLocalizationTest#doesNotRewritePlainStringMatchingADefaultNodeName`), and a node whose name is another type default is left alone (`#preservesDefaultNameOfADifferentNodeType`).

Residual, and inherent to any approach: a node the user deliberately renamed to exactly its own type default is indistinguishable from the default and still gets translated. That is fine. Resolving.

manager/tm/src/main/java/com/tapdata/tm/task/service/impl/TaskDagCheckLogServiceImpl.java
ply0011: Verified at e7acaf3a.

No `ThreadLocal` left in the locale path. `createLog` takes a `Locale`, and all three `MODEL_PROCESS_*` sites in `getLogs` pass the same `locale` the strategy loop uses; the node map is built with `localizeDagNodeName(locale, ...)` as well, so one response is now single-locale end to end.

`TaskDagCheckLogServiceImplLocalizationTest#dagCheckPassesLocaleExplicitlyToCreateLog` pins it (zh vs en in the same service instance, plus a `createLog` after `dagCheck` returns). Resolving.

manager/tm/src/main/resources/dagCheck_en_US.properties
ply0011: Verified at e7acaf3a. Key dropped.

Also checked the full key set rather than just this line: `NODE_NAME_*` in `dagCheck_en_US.properties` matches `NodeEnum` exactly, minus `cache_lookup_processor` — no remaining dead keys and no misspelled ones that would silently fall back. Resolving.

ply0011 (approved):

PR #3479, opened 2026-09-09
fix: TAP-12770 persist offsets after concurrent memory decrease

mnianqi:


iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNode.java
```diff
@@ -1606,12 +1628,12 @@ private void flushOffsetByTapdataEventForNoConcurrent(AtomicReference<TapdataEve
...
                 switch (syncStage) {
                     case INITIAL_SYNC:
-                        if (null != lastTapdataEvent.get().getBatchOffset() && !initialConcurrent) {
+                        if (null != lastTapdataEvent.get().getBatchOffset() && !isConcurrentProcessorRunning(syncStage)) {
                             flushSyncProgressMap(lastTapdataEvent.get());
                         }
                         break;
                     case CDC:
-                        if (null != lastTapdataEvent.get().getStreamOffset() && !cdcConcurrent) {
+                        if (null != lastTapdataEvent.get().getStreamOffset() && !isConcurrentProcessorRunning(syncStage)) {
```
ply0011: **correctness / high** — swapping the static `!initialConcurrent` / `!cdcConcurrent` guard for the live `!isConcurrentProcessorRunning(syncStage)` check opens an out-of-order offset flush on the partition-consumer threads.

`handleTapdataEvents` is the `eventProcessor` handed to `initCDCConcurrentProcessor` / `initInitialConcurrentProcessor` (lines 2237 / 2262 / 2290), so this method also runs on every partition thread of the concurrent processor — not just on the non-concurrent path. Previously that was harmless: `initialConcurrent` / `cdcConcurrent` are static config booleans, so a partition thread could never reach `flushSyncProgressMap`. Only the watermark runner advanced the offset, and only at a real watermark, i.e. after *every* partition had passed the barrier.

With the live check, the moment `currentRunning` flips to `false` the in-flight partition threads start flushing their **own** partition's last offset:

1. CDC concurrent mode, 3 partitions. Partition 2 throws on a target write.
2. `PartitionConcurrentProcessor.partitionConsumer` catches it and runs `currentRunning.compareAndSet(true, false)` *before* `errorHandler.accept(...)`.
3. Partitions 0 and 1 are still inside `eventProcessor.accept(processEvents)`. When they return, `isConcurrentProcessorRunning(CDC)` is now `false`, so each calls `flushSyncProgressMap(lastTapdataEvent.get())`.
4. Partition 0 is ahead at stream offset `T2`; partition 1 still has un-written events at `T1 < T2`. `T2` gets persisted, the task fails, and on restart CDC resumes at `T2` — the `T1` events are silently lost.

The same applies to the INITIAL_SYNC branch, and to the (narrower) window where `nodeRunning` goes false while a partition thread is between the per-event `isRunning()` loop and the flush.

The intent — flush when a DECREASE has parked the processor — is fine; it's the *probe* that's wrong, because it can't distinguish "this call came from the non-concurrent path" from "the concurrent processor just died under me". Suggest threading an explicit flag through instead, e.g. `handleTapdataEvents(events, boolean fromConcurrentProcessor)` (the concurrent processors pass `true`, `splitDDL2NewBatch`/`splitCompleteTableSnapshotEvent2NewBatch` pass `false`) and gating the flush on that rather than on live processor state.

iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/concurrent/PartitionConcurrentProcessor.java
```diff
@@ -184,15 +185,12 @@ protected void watermarkEventRunner() {
...
-						logger.info("waiting watermark event for all thread process, ts {}", sourceTime);
-					})) return; // when task stop, do not need flush offset
-					this.flushOffset.accept(event);
+				synchronized (watermarkFlushLock) {
+					if (!isRunning()) return;
+					final WatermarkEvent watermarkEvent = pollWatermarkEvent();
+					if (watermarkEvent != null && !flushWatermarkEvent(watermarkEvent)) {
+						return;
+					}
 				}
```
ply0011: **correctness / medium** — the monitor is held across the *blocking* poll, and the loop re-acquires it immediately, so `stop()` can be starved.

`pollWatermarkEvent()` is `watermarkQueue.poll(3, TimeUnit.SECONDS)`. Inside the `synchronized` block that means this thread owns `watermarkFlushLock` for a full 3 seconds on every idle iteration, then releases it and — via `while (isRunning())` — re-enters right away. Java monitors are barging, not FIFO, so a thread already running can win the re-acquire against one parked in `stop()`.

Why it matters: on a DECREASE, `handleTapdataAdjustMemoryEvent` calls `cdcPartitionConcurrentProcessor.stop()` on the Target-Queue-Consumer thread, while the source-facing `process()` thread is parked in `dynamicAdjustQueueLock.wait()` (`HazelcastTargetPdkBaseNode.java:988`) until the adjustment finishes. `stop()` gets past `waitingForProcessToCurrent()` and then blocks on `synchronized (watermarkFlushLock)` — every lost race is another 3 seconds with the whole target write path stalled.

Only the flush actually needs mutual exclusion; the poll doesn't. Either poll outside the lock and synchronize just around `flushWatermarkEvent`, or use a `ReentrantLock(true)` (fair) so `stop()` is guaranteed the next acquisition.

ply0011 (changes requested):
Code review (high) — 2 findings, details in the inline comments.

**High — `HazelcastTargetPdkBaseNode.java:1631/1636`: out-of-order offset flush from partition threads.** `handleTapdataEvents` is the `eventProcessor` of both concurrent processors, so `flushOffsetByTapdataEventForNoConcurrent` also runs on the partition-consumer threads. Replacing the static `!initialConcurrent` / `!cdcConcurrent` guard with the live `!isConcurrentProcessorRunning(syncStage)` check means that as soon as `currentRunning` flips false (e.g. one partition throws and `partitionConsumer` clears it before calling `errorHandler`), the other in-flight partition threads flush their *own* partition's last offset instead of a watermark. A partition that is ahead can persist an offset past events another partition never wrote → silent data loss on resume. Suggest an explicit "came from the non-concurrent path" flag rather than probing live processor state.

**Medium — `PartitionConcurrentProcessor.java:188`: `watermarkFlushLock` held across the blocking poll.** `pollWatermarkEvent()` blocks 3s inside the `synchronized` block and the loop re-acquires immediately; Java monitors are barging, so `stop()` can lose the race repeatedly. On a DECREASE that stalls the whole target write path (the source thread is parked in `dynamicAdjustQueueLock.wait()`) for a multiple of 3s. Poll outside the lock, or use a fair `ReentrantLock`.

The rest of the change reads correct — consolidating processor creation into `resumeInitialConcurrentProcessorIfNeed()` / `resumeCDCConcurrentProcessorIfNeed()` fixes the old INCREASE path, which called `initTargetConcurrentProcessorIfNeed()` and orphaned the still-running CDC processor by overwriting the field, and draining the watermark queue in `stop()` is the right place for the offset the DECREASE was dropping.

ply0011 (approved):

PR #3486, opened 2026-09-15
Tap 12830 2

jarad0628:


manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -189,7 +400,248 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+				}
+			} catch (Exception e) {
+				restartFailures.add("task " + taskId + " (" + stoppedTask.getName() + "): " + e.getMessage());
+				log.error("Failed to restart task {} after connector registration", taskId, e);
+			}
+		}
+	}
+
+	private boolean shouldRestartAfterRegistration(String statusBeforeStop, String statusNow) {
+		if (TaskDto.STATUS_STOP.equals(statusNow)) {
```
ply0011, line 712: **A task that was `scheduling` when it was selected is silently left stopped.**

`findAffectedTasks` snapshots `status` at query time, but `pause()` runs later. `scheduling` is a short-lived transient state: by the time `pauseAffectedTasks` calls `taskService.pause(...)`, the task has usually already moved to `wait_run`/`running`, so the state machine takes `WAITING_RUN/RUNNING --STOP--> STOPPING --> STOP` instead of `SCHEDULING --STOP--> SCHEDULING_FAILED`.

Restart then evaluates `shouldRestartAfterRegistration("scheduling", "stop")`:
- `statusNow == STOP` → returns `RUNNING.equals(before) || WAIT_RUN.equals(before)` → **false**
- `shouldReportRestartFailure("scheduling", "stop")` → false (not `error`)

So the task falls into the final `else`, which only emits a `log.warn`. Nothing is added to `restartFailures`, `uploadPdk` returns success, and a live business task stays permanently in `stop` with no error surfaced to the operator — exactly the failure mode the `restartFailures` machinery was added to prevent.

Suggest treating `scheduling` as an "active before stop" status for the `stop` case as well:

```suggestion
		if (TaskDto.STATUS_STOP.equals(statusNow)) {
			return TaskDto.STATUS_RUNNING.equals(statusBeforeStop)
					|| TaskDto.STATUS_WAIT_RUN.equals(statusBeforeStop)
					|| TaskDto.STATUS_SCHEDULING.equals(statusBeforeStop);
```

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -189,7 +400,248 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+				if (task == null) {
+					continue;
+				}
+				if (TaskDto.STATUS_STOPPING.equals(task.getStatus())) {
+					taskService.pause(taskId, user, false, true);
+				} else if (shouldRestartAfterRegistration(stoppedTask.getStatus(), task.getStatus())) {
+					// The affected heartbeat and shared CDC tasks are restored explicitly below.
+					// Disable TaskService's dependency orchestration to avoid starting an already
+					// scheduling dependency again while the business task is being restored.
+					taskService.start(task, user, "00");
```
ply0011: **`startFlag = "00"` skips shared-mining / heartbeat orchestration, which is not always covered by "restored explicitly below".**

In `TaskServiceImpl.start(taskDto, user, startFlag)`, bit 0 gates `logCollectorService.logCollector(user, taskDto)` and bit 1 gates `logCollectorService.startConnHeartbeat(user, taskDto)`. `logCollector()` is not just "wiring" — it looks up the existing shared-mining task and, in `newShareCdcProcess`/`oldShareCdcProcess`, calls `taskService.start(oldLogCollectorTask.getId(), user)` when that miner is **not** running (`tm-enterprise/.../LogCollectorServiceImpl.java`).

The comment assumes the miner and heartbeat tasks are always in `stoppedTasks` and restarted separately, but `findAffectedTasks` only selects tasks in `scheduling`/`wait_run`/`running`. Concrete scenario:

1. Task T has `shareCdcEnable = true`; its `logCollector` miner task is currently in `stop` (crashed / stopped earlier).
2. Registration selects T (running) but **not** the miner (status `stop`), stops T, uploads the definition.
3. T is restarted with `"00"` → `logCollector()` never runs → the miner is never started.
4. T comes back up with no shared-CDC source and stalls/errors.

Same shape for `connHeartbeat` when the heartbeat task was already stopped.

Either use `"11"` here (the orchestration is idempotent — it checks the miner's status before starting it), or include `stop`-state miner/heartbeat tasks in the restore set.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -189,7 +400,248 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+	}
+
+	private void restartAffectedInspects(List<InspectDto> stoppedInspects, UserDetail user, List<String> restartFailures) {
+		ListIterator<InspectDto> iterator = stoppedInspects.listIterator(stoppedInspects.size());
+		while (iterator.hasPrevious()) {
+			InspectDto stoppedInspect = iterator.previous();
+			ObjectId inspectId = stoppedInspect.getId();
+			try {
+				InspectDto inspect = inspectService.findById(inspectId);
+				if (inspect == null || !isInspectStopped(inspect.getStatus())) {
```
ply0011: **Inspects that did not reach a terminal state are skipped silently — no `restartFailures` entry.**

`isInspectStopped` only covers `done`/`passed`/`failed`/`error`. An inspect we set to `stopping` can still be in `stopping` (agent slow/offline) or back in `waiting` when the restore pass runs — in particular on every `waitForAffectedResourcesStopped` timeout, which is precisely the path where an inspect is stuck.

In that case this `continue` drops it: the inspect stays stopped, nothing is added to `restartFailures`, and nothing beyond the timeout message tells the operator that a running verification job was killed and never restarted. Tasks get the symmetric case handled (`STATUS_STOPPING` → `pause(..., restart=true)` sets the restart flag); inspects get nothing.

At minimum record it:

```suggestion
				if (inspect == null) {
					continue;
				}
				if (!isInspectStopped(inspect.getStatus())) {
					restartFailures.add("inspect task " + inspectId + " (" + stoppedInspect.getName()
							+ "): status is '" + inspect.getStatus() + "' and it was not restarted");
					continue;
				}
```

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -75,8 +110,184 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+		List<InspectDto> stoppedInspects = new ArrayList<>();
+		List<String> restartFailures = new ArrayList<>();
+		RuntimeException registrationFailure = null;
+		try {
+			Set<String> connectionIds = findAffectedConnectionIds(pdkSourceDtos, user);
+			List<TaskDto> affectedTasks = findAffectedTasks(connectionIds, user);
+			// Besides the per-connector locks acquired above, serialise registrations that touch the same
+			// tasks. Two different connectors may be referenced by one task, so connector level locks alone
+			// do not stop two registrations from stopping/restarting the same task concurrently.
+			acquireAffectedTaskLocks(affectedTasks, lockOwner, registrationLocks);
```
ply0011: **`registrationLocks` is mutated here while the renewal thread iterates the same `ArrayList`.**

`scheduleRegistrationLockRenewal(registrationLocks, ...)` was already called on line 119 and hands the *same* list to `RegistrationLockRenewal`, whose `renewAll()` runs on `DBLock.executor` and does `for (ILock lock : locks)`. This call then appends task locks to that list from the request thread with no synchronisation on the writer side (`renewAll` synchronises on `monitor`, but nothing here does).

Two consequences:

1. **CME kills renewal permanently.** If the append overlaps a renewal pass, the enhanced-for throws `ConcurrentModificationException` from `iterator.next()`, which is *outside* the per-lock `try/catch`. An uncaught exception from a `scheduleWithFixedDelay` task cancels the future for good — no lock is ever renewed again, the 300 s lease expires mid-registration, and a concurrent registration takes over (`MongoDBLockRepository.renew` grants an expired lock to a new owner). The window is not purely theoretical: `findAffectedTasks` runs an unindexed `dag.nodes.connectionId` scan over the whole task collection, and the first renewal fires 60 s in.
2. **Visibility.** Locks appended after the renewal thread started have no happens-before edge to `renewAll`, so the task locks may never be renewed at all and silently expire after 300 s.

Cheapest fix: make the field a `CopyOnWriteArrayList` (or have `acquireAffectedTaskLocks` publish through the renewal object under `monitor`).

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -189,7 +400,248 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+		List<Criteria> connectionCriteria = new ArrayList<>();
+		connectionCriteria.add(Criteria.where("dag.nodes.connectionId").in(connectionIds));
+		connectionCriteria.add(Criteria.where("dag.nodes.connectionIds").in(connectionIds));
+		connectionIds.forEach(connectionId -> connectionCriteria.add(
+				Criteria.where("dag.nodes.logCollectorConnConfigs." + connectionId).exists(true)));
+
+		Criteria criteria = new Criteria().andOperator(
+				Criteria.where("is_deleted").ne(true),
+				Criteria.where("syncType").in(TaskDto.SYNC_TYPE_SYNC, TaskDto.SYNC_TYPE_MIGRATE,
+						TaskDto.SYNC_TYPE_LOG_COLLECTOR, TaskDto.SYNC_TYPE_CONN_HEARTBEAT),
```
ply0011: **`mem_cache` (shared cache) tasks are missing from the affected set.**

`TaskDto.SYNC_TYPE_MEM_CACHE` tasks are ordinary engine jobs that read a source through a PDK connector (`LogCollectorServiceImpl.logCollector` explicitly handles `SYNC_TYPE_MEM_CACHE` alongside `sync`/`migrate`). A running shared-cache task over a connection of the connector being registered keeps the old jar loaded, which is the exact condition this PR exists to eliminate — and since it is never stopped, it is also never restarted, so it silently keeps serving stale connector behaviour after the registration reports success.

Either add `TaskDto.SYNC_TYPE_MEM_CACHE` to this `in(...)` list, or state in a comment why shared-cache tasks are deliberately exempt.

manager/tm/src/main/java/com/tapdata/tm/ds/controller/PdkController.java
```diff
@@ -55,7 +57,10 @@ public ResponseMessage<Void> uploadJar(@RequestPart(value = "file") MultipartFil
             pdkSourceDtos.add(pdkSourceDto);
         }
 
-        pkdSourceService.uploadPdk(file, pdkSourceDtos, latest, getLoginUser());
+        UserDetail user = getLoginUser();
+        boolean accessCodeRegistration = user != null
```
ply0011, line 63: **`authType` is only populated on one of four authentication paths, so this flag fails dangerous.**

`LoginUserResolver.doResolve` sets `userDetail.setAuthType(...)` **only** in the `access_token` query-parameter branch. The other three branches — `specifiedUserId`, the `user_id` request header, and HTTP `Basic` auth — return a `UserDetail` with `authType == null`.

For any of those callers `accessCodeRegistration` evaluates to `false`, which selects the destructive branch: every `running`/`wait_run`/`scheduling` task and inspect touching that connector is stopped, waited on for up to 60 s, and restarted. A caller that previously just uploaded a definition now takes down live tasks, and the failure is silent in the sense that nothing in the request distinguishes "this is an access-code registration" from "we could not tell".

Given the blast radius, the safe default is the opposite one: treat an unknown/absent `authType` as an access-code registration (skip the stop/restart), or populate `authType` on the other resolver branches so the decision is always made on real data.

tapdata-cli/src/main/java/io/tapdata/pdk/cli/services/UploadFileService.java
```diff
@@ -39,37 +51,22 @@
...
-
-      Map map = JSON.parseObject(s, Map.class);
-      Object data = map.get("data");
-      if (null == data) {
-        printUtil.print(PrintUtil.TYPE.ERROR, "TM sever not found or generate token failed");
-        return;
-      }
-      JSONObject data1 = (JSONObject) data;
-      token = (String) data1.get("id");
+      token = StringUtils.isNotBlank(accessCode)
```
ply0011, line 66: **Access-code token failure returns silently, and the caller then prints "Completed" for a jar that was never uploaded.**

`generateAccessCodeToken` returns `null` on failure, so the `if (StringUtils.isBlank(token)) return;` below leaves `upload()` before the HTTP request is made. `RegisterCli:255-256` does not inspect a return value — it prints `"* Register Connector: %s | (%s) Completed"` immediately after. Running `register -a <bad-or-expired-accesscode> -t <tm> *.jar` therefore reports every connector as registered while nothing was uploaded, and the CLI exits 0.

The new `login()` path is inconsistent with this: it throws `IllegalStateException` on the same class of failure. Making `generateAccessCodeToken` throw as well (instead of `return null`) would remove the early-return branch entirely and give both paths the same behaviour.

ply0011 (changes requested):
Reviewed the connector-registration stop/restart flow (high effort). 7 findings, posted inline.

**High**
- `PkdSourceService#shouldRestartAfterRegistration` — a task observed as `scheduling` that advances to `wait_run`/`running` before `pause()` lands ends in `stop`, which is neither restarted nor reported. It stays down permanently and the registration reports success.
- `PkdSourceService#restartAffectedTasks` — restarting with `startFlag "00"` skips `logCollector()`/`startConnHeartbeat()`. A shared-mining task that was already in `stop` is not in the affected set and is never started, so a `shareCdcEnable` task comes back without a CDC source.

**Medium**
- `restartAffectedInspects` silently skips inspects still in `stopping` without recording a restart failure (tasks have the symmetric path, inspects don't).
- `registrationLocks` is appended to by `acquireAffectedTaskLocks` while `RegistrationLockRenewal` iterates the same `ArrayList` — a CME there permanently cancels the `scheduleWithFixedDelay` future and the 300 s lease then expires mid-registration.
- `findAffectedTasks` omits `SYNC_TYPE_MEM_CACHE`, so running shared-cache tasks keep the old jar loaded.
- `PdkController` — `authType` is only populated on the `access_token` branch of `LoginUserResolver`; every other auth path yields `null` and therefore selects the stop-all-tasks branch by default.

**Low**
- `UploadFileService.upload` returns silently when the access-code token cannot be obtained, and `RegisterCli` then prints "Completed" for a jar that was never uploaded.

HarsenLin:

- Status: **ERROR**, MetricKey: **new_reliability_rating**, ActualValue: **2**


See Sonar Scan Result at: http://fileserver.tapdata.io:29000/dashboard?branch=TAP-12830-2&id=tapdata

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Verified at eb73b0a8: fixed in dc44b3ab. `shouldRestartAfterRegistration` now treats `scheduling` → `stop` as active, and `testRestartSchedulingTaskThatStoppedDuringRegistration` covers it. Resolving.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Re-checked at eb73b0a8. Switching to `"11"` fixes the case described above (the miner was already stopped before the registration), but it breaks the common case. The premise in my suggestion was wrong: the shared-mining orchestration is only idempotent when the miner is already `running`.

`newShareCdcProcess` / `oldShareCdcProcess` (tm-enterprise) return early only on `STATUS_RUNNING`. For any other status they call `taskService.start(minerId, user)` again, and nothing up the chain catches the exception:
- https://github.com/tapdata/tapdata-enterprise/blob/7a6b7753259221700e35dfe47d808f6e893f2e7b/tm-enterprise/src/main/java/com/tapdata/tm/task/service/LogCollectorServiceImpl.java#L1398-L1418
- https://github.com/tapdata/tapdata-enterprise/blob/7a6b7753259221700e35dfe47d808f6e893f2e7b/tm-enterprise/src/main/java/com/tapdata/tm/task/service/LogCollectorServiceImpl.java#L1473-L1494

`TaskServiceImpl.start` throws `Task.StartStatusInvalid` when the status is not in `to_start_status` (edit / stop / complete / error / schedule_failed / wait_start):
- https://github.com/tapdata/tapdata/blob/eb73b0a8cd97361c4b539d3d792a7a9a22d64def/manager/tm/src/main/java/com/tapdata/tm/task/service/TaskServiceImpl.java#L4685-L4691
- https://github.com/tapdata/tapdata/blob/eb73b0a8cd97361c4b539d3d792a7a9a22d64def/manager/tm/src/main/java/com/tapdata/tm/task/constant/TaskOpStatusEnum.java#L15

Failure path when the registered connector backs a shared-mining source connection C:
1. Task B (`shareCdcEnable`), its miner M and its heartbeat task H all reference C, so all three are paused.
2. The restore runs in reverse stop order: H → M → B. After `start(M, "11")`, M is `scheduling` / `wait_run` (the engine still has to fetch the newly registered jar).
3. `start(B, "11")` → `logCollector()` → M is not `running` → `start(M)` → `Task.StartStatusInvalid`. B never reaches `run()` and stays in `stop`, and the registration fails with "could not be restarted".

With `"00"` this case worked, because M was restored on its own. The heartbeat path has the same `!RUNNING → start` check, but it is wrapped in try/catch, so it only writes a spurious WARN to the task log ([L1574-L1602](https://github.com/tapdata/tapdata-enterprise/blob/7a6b7753259221700e35dfe47d808f6e893f2e7b/tm-enterprise/src/main/java/com/tapdata/tm/task/service/LogCollectorServiceImpl.java#L1574-L1602)). This only affects the enterprise build; the OSS `logCollector()` is a no-op. `PkdSourceServiceTest` mocks `TaskService`, so the tests cannot catch it. The same applies to shared-cache (`mem_cache`) tasks with shared mining enabled.

Either of these would fix it:
- Before restoring the consumers with `"11"`, wait until the miners restored in this pass reach `running` (poll like `waitForAffectedResourcesStopped`, and record a restart failure on timeout).
- Or use `"00"` for tasks whose miner / heartbeat task was restored in this pass, and `"11"` only when the dependency is not in the restore set (the case this thread originally described).

Leaving this thread open.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Verified at eb73b0a8: fixed in 1c1e1ee5. `restartAffectedInspects` now adds both a missing inspect and an inspect that is not in a terminal state to `restartFailures`. No unit test covers this path yet. Resolving.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Verified at eb73b0a8: fixed. The lock list is now a `CopyOnWriteArrayList` shared by reference with `RegistrationLockRenewal`, so the appends from `acquireAffectedTaskLocks` / `acquireAffectedInspectLocks` are safe and visible to the renewal thread (c35b9a32, 1c1e1ee5). A failed renewal is now recorded, and the registration aborts at the next `ensureHealthy()` checkpoint (1c1e1ee5, eb73b0a8). The renewal-failure abort path has no test yet. Resolving.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Verified at eb73b0a8: fixed in c35b9a32. `SYNC_TYPE_MEM_CACHE` is now in the `findAffectedTasks` filter. Shared-cache tasks with shared mining enabled hit the same restore failure described in the startFlag thread. Resolving.

manager/tm/src/main/java/com/tapdata/tm/ds/controller/PdkController.java
ply0011: Verified at eb73b0a8: fixed in 1c1e1ee5. A missing or unknown `authType` now selects the upload-only path, and the other three resolver branches set `authType` explicitly.

One thing to confirm: the `user_id` header branch (cloud gateway, i.e. CLI `-ak/-sk`) and the Basic branch are labelled `USERNAME_LOGIN`, so cloud registrations now take the stop/restart path by design. Please confirm that is intended. There is also no `PdkController` test for this decision. Resolving.

tapdata-cli/src/main/java/io/tapdata/pdk/cli/services/UploadFileService.java
ply0011: Verified at eb73b0a8: fixed. `generateAccessCodeToken` now throws, and `upload()` throws on a blank token and on a non-`ok` response, so `RegisterCli` exits -1 instead of printing "Completed" (c35b9a32, 59b5e248).

Side effect: in directory mode, one failed jar now aborts the rest of the batch. `build/image/docker-entrypoint.sh` registers `connectors/dist` in a single call and exits the container on failure. The official connectors are all `1.0-SNAPSHOT`, so normal restarts are unaffected, but a non-SNAPSHOT custom jar left in `dist` would now block startup (TM rejects re-registering that version). Resolving.

ply0011 (changes requested):
Re-verified the 7 findings at eb73b0a8.

**Fixed and resolved (6):** `scheduling` → `stop` restore, inspect restart failures, lock-list CME (plus fail-fast renewal), `mem_cache` in the affected set, the `authType` default, and CLI failure reporting. Locally `PkdSourceServiceTest` (31), `LoginUserResolverTest` (14) and `PdkControllerTest` (2) all pass, and Scan-Tapdata is green on this SHA.

**Still blocking (1):** restoring with startFlag `"11"` breaks the common shared-mining case. The business task's `logCollector()` starts the miner again while the miner restored in the same pass is still `scheduling` / `wait_run`, and `start()` throws `Task.StartStatusInvalid`. The business task stays in `stop` and the registration fails. Details and two fix options are in the startFlag thread.

**Non-blocking:**
- Please confirm that cloud (`user_id` header, CLI `-ak/-sk`) and Basic registrations are meant to stop and restart tasks. No `PdkController` test covers this decision.
- CLI: one failed jar now aborts a directory-mode batch, and `build/image/docker-entrypoint.sh` then exits the container.
- `stoppedInspects.add()` now runs before `doExecuteInspect()`. If that call throws before changing the status, an inspect that is still running gets reported as "remained in status 'running'".
- The root `register_and_start.sh` calls `register` without `-a` / `-p`. With the new `validateAuthentication` it now fails, and `set -e` stops the engine from starting. Before this PR it silently registered nothing.
- No tests yet for: an inspect that did not stop, `mem_cache` in the query, the renewal-failure abort, the `findCurrentAffectedTasks` re-read, and the `PdkController` auth decision.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
```diff
@@ -189,7 +461,386 @@ public void uploadPdk(MultipartFile[] files, List<PdkSourceDto> pdkSourceDtos, b
...
+				|| InspectStatusEnum.PASSED.getValue().equals(status)
+				|| InspectStatusEnum.FAILED.getValue().equals(status)
+				|| InspectStatusEnum.ERROR.getValue().equals(status);
+	}
+
+	private boolean isInspectInProgress(String status) {
+		return InspectStatusEnum.RUNNING.getValue().equals(status)
+				|| InspectStatusEnum.SCHEDULING.getValue().equals(status)
+				|| InspectStatusEnum.WAITING.getValue().equals(status)
+				|| InspectStatusEnum.STOPPING.getValue().equals(status);
```
ply0011: **`STOPPING` here brings back the silently-dropped inspect case that was resolved in the previous review.**

In `restartAffectedInspects`, any status in `isInspectInProgress` only logs a WARN and `continue`s. That is fine for `running` / `scheduling` / `waiting`, because the inspect is up: either it never stopped or someone restarted it. `stopping` is different. The inspect is on its way down and nothing will bring it back:

- `waitForAffectedResourcesStopped` times out, so the restore runs while the inspect is still `stopping`.
- `stopAffectedInspects` fails partway, so the restore runs immediately and every inspect stopped before the failure is still `stopping`.

In both cases the inspect ends in `done` / `error`, is never rescheduled, and is not in `restartFailures`. A task in the same position gets `pause(..., restart=true)`; an inspect gets nothing.

Moving `stoppedInspects.add()` after `doExecuteInspect()` widens this. In tm-enterprise, `doExecuteInspect(STOPPING)` calls `update()` (status → `stopping`) *before* `inspectTaskStop()`. If `inspectTaskStop()` throws, the inspect is `stopping` in the DB but is not in `stoppedInspects`, so the restore never looks at it.

`testDoNotReportRunningInspectAsRestartFailure` only covers `running`.

Suggest dropping `STOPPING` so the inspect reaches the `restartFailures` branch. Alternatively, wait for it the way `waitForTaskReadyToRestart` does for tasks, then reschedule it:

```suggestion
				|| InspectStatusEnum.WAITING.getValue().equals(status);
```

ply0011 (changes requested):
Re-verified at d886a585, together with tapdata/tapdata-enterprise#857 (6c312f41).

**Fixed (1):** the startFlag `"11"` / shared-mining restore. #857 treats `scheduling` / `wait_run` miners and heartbeat tasks as already started, and this PR skips entries that are already active and retries on `StartStatusInvalid`. **This holds only if #857 merges first or together with this PR.** Without #857 the enterprise build still fails B's restore; details are in the startFlag thread.

**Previous non-blocking items:** all addressed. Cloud / `user_id` / Basic registrations now upload only, the CLI attempts every jar before failing, `register_and_start.sh` skips registration when no credentials are set, and the five missing tests were added.

**Blocking (1):** `isInspectInProgress` includes `STOPPING`, which reopens the resolved inspect finding (inline on L665). The fix is one line.

**Non-blocking, please confirm intent:**
- `shouldRestartAfterRegistration` now restarts a task that went from active to `error` during the stop instead of reporting it. A persistent error then shows up only after the registration has already reported success.
- `register_and_start.sh` with only `TAPDATA_PASSWORD` set uses `-u/-p` (`USERNAME_LOGIN`), so every container start takes the stop/restart path. That contradicts the "interactive only" intent in `PdkController`. Consider using access-code only in the startup script.
- #857 `pause()`: a `scheduling` miner is now paused into `schedule_failed`, where `restartFlag` has no effect because `TaskRestartSchedule.restartTask` only picks up `stop`. The miner comes back through `retrySchedulingFailedTask` after about 60 s, and only on non-cloud deployments.

**Verification:** local runs at d886a585:
- `PkdSourceServiceTest` (37), `PdkControllerTest` (12) and `LoginUserResolverTest` (14) pass.
- `iengine-common`, `validator` and `iengine-app` compile against 2.0.11 / 2.11-SNAPSHOT (tapdata/tapdata-common-lib#212). `PdkUtilTest` (12) and `ConnectionNodeUtilTest` (5) pass.
- `LogCollectorServiceTest` (70) passes on #857.

CI has not run on this head: d886a585 has no check runs, Scan-Tapdata on 3c6594e7 is still queued, and the last Sonar gate (09-21) was ERROR. Please re-trigger it.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Re-verified at d886a585 together with tapdata/tapdata-enterprise#857 (6c312f41). Fixed, provided both PRs land together.

- #857 makes `newShareCdcProcess` / `oldShareCdcProcess` / `startConnHeartbeat` treat `scheduling` / `wait_run` as already started. `start(B, "11")` → `logCollector()` therefore no longer calls `start(M)` while the miner restored earlier in the same H → M → B pass is still scheduling.
- This PR skips entries that are already active when the restore pass reaches them, and `restartTaskById` reloads by id and retries on `Task.StartStatusInvalid`.

This differs from both options I suggested, and it is the better fix: it also covers the non-registration case where two consumers of one miner start concurrently.

Merge order matters. With this PR alone (no #857), the enterprise build still fails here: the nested `start(M)` throws `StartStatusInvalid`, `restartTaskById` sees B in `stop` and retries three times back to back with no wait while M is still `scheduling`, and B ends up in `restartFailures`. Please merge #857 first or together. `PkdSourceServiceTest` mocks `TaskService`, so this path is verified by reading the code, not by a test. Resolving.

manager/tm/src/main/java/com/tapdata/tm/ds/service/impl/PkdSourceService.java
ply0011: Reopening. dcf079cd brings this case back: `isInspectInProgress` includes `STOPPING`, so an inspect that is still `stopping` when the restore runs is again skipped with only a WARN. It is not restarted and not added to `restartFailures`. Details are in the new comment on `isInspectInProgress`.

PR #3499, opened 2026-09-18
fix: TAP-12770 persist offsets after concurrent memory decrease

mnianqi:


iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNode.java
```diff
@@ -2081,6 +2081,12 @@ protected static void ignorePksAndIndices(TapTable tapTable, List<String> logicP
 
     @Override
     public boolean saveToSnapshot() {
+        synchronized (this.saveSnapshotLock) {
```
ply0011: **`saveSnapshotLock` now spans two remote calls, and `doClose()` waits on it uninterruptibly.**

Moving the whole body of `saveToSnapshot()` under `saveSnapshotLock` means the monitor is now held across `flushOffsetFunction.flushOffset(...)` for every entry in `sourceConnectorNodeMap` and across `clientMongoOperator.insertOne(syncProgressJsonMap, "task/syncProgress/<id>")` — a REST call to TM.

Concrete failure: the target has `offsetCallbackEnable = true`, so `initFlushOffsetConsumer` makes the connector's own callback thread call `saveToSnapshot()` on *every* offset callback. TM is restarting, so `insertOne` sits on its HTTP socket timeout while that callback thread holds the lock. The user stops the task:

```
doClose()
  -> stopFlushOffsetExecutorAndSaveSnapshot()
     -> flushOffsetExecutor.shutdownNow()   // interrupts only the Flush-Offset thread
     -> saveToSnapshot()                    // synchronized(saveSnapshotLock) -> parks
```

`synchronized` is not interruptible and nothing interrupts the connector callback thread, so the close thread parks for the whole remote timeout. Before this change doClose only paid for its own save; now it pays for someone else's first. The same applies to the Target-Queue-Consumer thread inside `handleTapdataStartCdcEvent` — `queueConsumerThreadPool.shutdownNow()` at L2340 cannot pull a thread out of a monitor wait.

Worth narrowing the critical section to what actually needs serializing (build `syncProgressJsonMap` under the lock, do the `flushOffset` / `insertOne` calls outside), or using a `ReentrantLock` with `tryLock(timeout)` on the close path so a stuck peer cannot hold task shutdown hostage.

iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNode.java
```diff
@@ -2081,6 +2081,12 @@ protected static void ignorePksAndIndices(TapTable tapTable, List<String> logicP
...
+            return saveToSnapshotLocked();
+        }
+    }
+
+    private boolean saveToSnapshotLocked() {
```
ply0011: **The new lock only guards the reader; the `SyncProgress` mutators are still unsynchronized, so the snapshot can still be torn.**

`saveSnapshotLock` now serializes `saveToSnapshot()` against itself, but `flushOffsetCallback` / `flushSyncProgressMap` mutate the *same* `SyncProgress` instances from the queue-consumer, watermark and connector-callback threads without ever taking it. `saveToSnapshotLocked` reads those mutable objects field by field and then serializes the whole object.

Concrete failure:

1. Flush-Offset-Thread enters `saveToSnapshotLocked`, reaches `syncProgress.setBatchOffset(PdkUtil.encodeOffset(syncProgress.getBatchOffsetObj()))` and encodes the batch offset as of T1.
2. Before it reaches the `getStreamOffsetObj()` block a few lines down, the watermark thread runs `flushSyncProgressMap` -> `flushOffsetCallback` and does `syncProgress.setStreamOffsetObj(newer)` (T2).
3. The flush thread encodes *that* newer stream offset and `JSONUtil.obj2Json(syncProgress)` persists `{batchOffset: T1, streamOffset: T2}`.

On restart the task resumes the snapshot phase from the older batch offset but starts CDC from the newer stream offset, so the CDC events between T1 and T2 are never replayed — silent data loss, which is exactly the class of problem this ticket is about.

Note the same block reads `syncProgress.getStreamOffsetObj()` three times (null check, `instanceof String` check, `encodeOffset` argument); if it flips between reads you can also double-encode an already-encoded offset, since the `STREAM_OFFSET_COMPRESS_PREFIX_V2` / `ENCODE_PREFIX` guard was evaluated against a different value.

Either take `saveSnapshotLock` in `flushOffsetCallback` / `flushSyncProgressMap` too, or read each `SyncProgress` into locals once under the lock and serialize from those.

iengine/iengine-app/src/test/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNodeTest.java
```diff
@@ -1078,6 +1078,50 @@ void testMainProcess() {
...
+					assertFalse(future.isDone());
+				}
+				assertTrue(future.get(1, TimeUnit.SECONDS));
+			} finally {
+				executor.shutdownNow();
+			}
+		}
+
+		@Test
+		void closeStopsScheduledSnapshotsBeforeTheFinalSnapshot() {
```
ply0011: **This test never calls `doClose()`, so it does not cover the regression the PR fixes.**

It reflects straight into the private helper `stopFlushOffsetExecutorAndSaveSnapshot`, so the only thing asserted is that the helper calls `shutdownNow()` before `saveToSnapshot()` — which is visible in three lines of the helper itself. Nothing ties that helper to `doClose()`.

Delete or revert L2346 (`CommonUtils.ignoreAnyError(this::stopFlushOffsetExecutorAndSaveSnapshot, TAG)`) back to the old pair, or drop the final save from `doClose()` altogether, and this test still passes green.

Calling the real `doClose()` (with `exactlyOnceWriteCleanerEntities` empty and the other collaborators stubbed) and asserting the `shutdownNow` → `saveToSnapshot` order there would actually pin the behaviour.

ply0011 (changes requested):
Code review (high effort) — 3 findings, 2 of them on the production change.

**Blocking**

1. `HazelcastTargetPdkBaseNode.java:2084` — widening `saveSnapshotLock` to the whole `saveToSnapshot()` body puts the monitor around `flushOffset()` on every source connector *and* the `insertOne` REST call to TM. `doClose()` now waits on that monitor, `synchronized` is not interruptible, and `flushOffsetExecutor.shutdownNow()` does not touch the connector offset-callback thread or the queue-consumer thread that also call `saveToSnapshot()`. A slow/hung TM turns task stop into a park for the full remote timeout.

2. `HazelcastTargetPdkBaseNode.java:2089` — the lock only covers the reader. `flushOffsetCallback` / `flushSyncProgressMap` still mutate the same `SyncProgress` instances from other threads without it, so `saveToSnapshotLocked` can serialize `{batchOffset: T1, streamOffset: T2}` and skip the CDC events in between on restart. The race this ticket is about is only half closed.

**Non-blocking**

3. `HazelcastTargetPdkBaseNodeTest.java:1111` — `closeStopsScheduledSnapshotsBeforeTheFinalSnapshot` reflects into the private helper and never calls `doClose()`, so reverting the `doClose()` line still leaves the suite green.

Details in the inline comments.

iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNode.java
ply0011: 已在 `cb86c433` 修复，复核通过。

改法：抽出 `createSnapshotPayload()`，`saveSnapshotLock` 现在只覆盖「读 `syncProgressMap` + 编码 + 拼 JSON」这段纯 CPU 工作；`flushOffsetFunction.flushOffset(...)` 和 `clientMongoOperator.insertOne(...)` 都移到锁外执行，uploadDag 那段换成独立的 `uploadDagLock`。

复核要点：
- 并发前提确认成立 —— `saveToSnapshot()` 的四个调用方分属不同线程（10s 周期的 `flushOffsetExecutor`、`initFlushOffsetConsumer` 注册进 `setFlushOffsetCallback` 的连接器回调线程、`handleTapdataStartCdcEvent` 所在的 Target-Queue-Consumer、`doClose()`），所以原先的阻塞路径是真实的。
- 对照 develop：`saveSnapshotLock` 原本只护 uploadDag 一小段，`saveToSnapshot` 本身无锁，这次阻塞确属 e45d4a30 新引入，现已消除。
- `HazelcastTargetPdkBaseNodeTest` 全类 141 tests 全绿。

iengine/iengine-app/src/main/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNode.java
ply0011: 已在 `cb86c433` 修复，复核通过。建议的两条路作者都走了。

改法：`flushOffsetCallback` 与 `flushSyncProgressMap` 都加上同一把 `saveSnapshotLock`；同时 `createSnapshotPayload()` 里先 `copySyncProgress(entry.getValue())` 复制一份再编码、序列化，所以持久化的是锁内取到的一致快照。附带提到的三次重复读也修了 —— `streamOffsetObj` 现在收进局部变量只读一次，不会再出现「守卫判断和 `encodeOffset` 看到不同值」的重复编码。

复核要点：
- 逐字段核对 `copySyncProgress`：`SyncProgress` 在本 PR base 上的 11 个字段全部复制，无遗漏。
- 全量 grep 写侧，剩余两处 `syncProgress.set*` 在 `initSyncProgressMap` / `readBatchOffset(SyncProgress)` 内，都在 `doInit` 排定调度器之前执行，单线程，不构成漏网。
- Mutation 验证：去掉 `flushOffsetCallback` 上的 `synchronized` 后 `progressUpdatesUseTheSnapshotLock` 失败（`expected: <false> but was: <true>`），说明新用例确实咬住了这把锁。

iengine/iengine-app/src/test/java/io/tapdata/flow/engine/V2/node/hazelcast/data/pdk/HazelcastTargetPdkBaseNodeTest.java
ply0011: 已在 `cb86c433` 修复，复核通过。

改法：`closeStopsScheduledSnapshotsBeforeTheFinalSnapshot` 改为调用真实的 `doClose()`（`exactlyOnceWriteCleanerEntities` 为空、其余协作者打桩），再 `inOrder` 验 `shutdownNow()` → `saveToSnapshot()`。

Mutation 验证：把 `doClose()` 里那行换回「只 `shutdownNow`、不 save」，该用例失败（HazelcastTargetPdkBaseNodeTest.java:1185）—— 现在确实能钉住行为。

顺带一条复核时发现、原评论没提的：`doClose()` 这部分改动相对 develop **在行为上是零**。develop 上本来就是

```java
CommonUtils.ignoreAnyError(() -> Optional.ofNullable(this.flushOffsetExecutor).ifPresent(ExecutorService::shutdownNow), TAG);
CommonUtils.ignoreAnyError(this::saveToSnapshot, TAG);
```

顺序已经是 `shutdownNow` → `saveToSnapshot`，本 PR 只是把两行抽成 helper。也就是说这个 PR 真正的功能性修复只有加锁那一块，关闭顺序本身 develop 上已经是对的。另外 helper 丢掉了原来的 `Optional.ofNullable` null 保护，不过 `flushOffsetExecutor` 是 `final` 且构造器赋值，生产上不会为 null，仅对 mock 测试有影响。

ply0011 (approved):

